PR #7, opened 2017-09-03
modified php settings

bradymiller:
per settings here(not kept memory still at 128 since low resource environment):
http://www.open-emr.org/wiki/index.php/FAQ#What_are_the_correct_PHP_settings_.28can_be_found_in_the_php.ini_file.29.3F


bradymiller:
@TheToolbox ,
If you're ok with this, then feel free to merge it in.
-brady

TheToolbox:
LGTM!

PR #14, opened 2017-09-27
Reorganization

jesdynf:


jesdynf:
@toolbox reorg: bring Docker images into primacy, out of Lightsail's umbrella

PR #15, opened 2017-10-02
File Permission Hardening

TheToolbox:
This adds significant security improvements and hopefully should make the docker image ready for prime-time. The commits should explain themselves!

### DIFF
--- a/docker-images/openemr/Dockerfile
+++ b/docker-images/openemr/Dockerfile
@@ -16,6 +16,7 @@ VOLUME [ "sites" ]
 ENV APACHE_LOG_DIR=/var/log/apache2
 COPY php.ini /etc/php7/php.ini
 COPY openemr.conf /etc/apache2/conf.d/
+#add runner and auto_configure and prevent auto_configure from being run w/o being enabled
 COPY run_openemr.sh .
 COPY auto_configure.php .
 RUN chmod 700 run_openemr.sh \
@@ -23,8 +24,6 @@ RUN chmod 700 run_openemr.sh \
     && ln -s /usr/bin/php7 /usr/bin/php
 #fix issue with apache2 dying prematurely
 RUN mkdir -p /run/apache2
-#add user to run apache w/ reduced permissions
-RUN adduser -D www && chown -Rf www:www . && chown -Rf www:www /var/log/apache2
 #go
 CMD [ "./run_openemr.sh" ]
 
--- a/docker-images/openemr/Dockerfile
+++ b/docker-images/openemr/Dockerfile
@@ -11,7 +11,7 @@ RUN git clone https://github.com/openemr/openemr.git --branch rel-500 --depth 1 
     && chown -R apache openemr/ \
     && mv openemr /var/www/localhost/htdocs/
 WORKDIR /var/www/localhost/htdocs/openemr
-VOLUME [ "sites" ]
+VOLUME [ "/var/www/localhost/htdocs/openemr/sites" ]
 #configure apache & php properly
 ENV APACHE_LOG_DIR=/var/log/apache2
 COPY php.ini /etc/php7/php.ini
--- a/docker-images/openemr/Dockerfile
+++ b/docker-images/openemr/Dockerfile
@@ -10,17 +10,22 @@ RUN git clone https://github.com/openemr/openemr.git --branch rel-500 --depth 1 
     && chmod 666 openemr/interface/modules/zend_modules/config/application.config.php \
     && chown -R apache openemr/ \
     && mv openemr /var/www/localhost/htdocs/
+WORKDIR /var/www/localhost/htdocs/openemr
+VOLUME [ "sites" ]
 #configure apache & php properly
+ENV APACHE_LOG_DIR=/var/log/apache2
 COPY php.ini /etc/php7/php.ini
 COPY openemr.conf /etc/apache2/conf.d/
 COPY run_openemr.sh .
-RUN chmod 700 run_openemr.sh
 COPY auto_configure.php .
-RUN ln -s /usr/bin/php7 /usr/bin/php
+RUN chmod 700 run_openemr.sh \
+    && chmod 000 auto_configure.php \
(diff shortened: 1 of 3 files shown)


docker-images/openemr/Dockerfile
```diff
@@ -10,17 +10,22 @@ RUN git clone https://github.com/openemr/openemr.git --branch rel-500 --depth 1
     && chmod 666 openemr/interface/modules/zend_modules/config/application.config.php \
     && chown -R apache openemr/ \
     && mv openemr /var/www/localhost/htdocs/
+WORKDIR /var/www/localhost/htdocs/openemr
+VOLUME [ "sites" ]
```
jesdynf: Should this VOLUME command have a full path? Have you verified it's pointing where you want? The docs list the commands that WORKDIR sets the local path to, and VOLUME isn't on that list...

docker-images/openemr/run_openemr.sh
```diff
@@ -1,9 +1,53 @@
...
+    #set all directories to 500
+    find . -type d -print0 | xargs -0 chmod 500
+    #set all file access to 400
+    find . -type f -print0 | xargs -0 chmod 400
+
+    echo "Default file permissions and ownership set, allowing writing to specific directories"
+    chmod 700 run_openemr.sh
+    # Set file and directory permissions
+    #chmod 666 sites/default/sqlconf.php why?
+    chmod 600 interface/modules/zend_modules/config/application.config.php # should this actually be writable?
```
bradymiller: yes, this needs to be writable (annoying zend...)

docker-images/openemr/run_openemr.sh
```diff
@@ -1,9 +1,53 @@
...
+    find sites/default/documents -type d -print0 | xargs -0 chmod 700
+    find sites/default/edi -type d -print0 | xargs -0 chmod 700
+    find sites/default/era -type d -print0 | xargs -0 chmod 700
+    find sites/default/letter_templates -type d -print0 | xargs -0 chmod 700
+    find interface/main/calendar/modules/PostCalendar/pntemplates/cache -type d -print0 | xargs -0 chmod 700
+    find interface/main/calendar/modules/PostCalendar/pntemplates/compiled -type d -print0 | xargs -0 chmod 700
+    find gacl/admin/templates_c -type d -print0 | xargs -0 chmod 700
```
bradymiller: Just make sure this find/xarg stuff will apply to stuff that gets put in these places by openemr (for example, when openemr places a new document in sites/default/documents/1/<some file> ).
Also maybe a stupid question, but why 700 and not 600 for above stuff?

bradymiller: I think i figured out that you did 700 instead of 600 since are directories :)

docker-images/openemr/run_openemr.sh
```diff
@@ -1,9 +1,53 @@
...
+    echo "OpenEMR configured. Setting user 'www' as owner of openemr/ and setting file/dir permissions to 400/500"
+    #set all directories to 500
+    find . -type d -print0 | xargs -0 chmod 500
+    #set all file access to 400
+    find . -type f -print0 | xargs -0 chmod 400
+
+    echo "Default file permissions and ownership set, allowing writing to specific directories"
+    chmod 700 run_openemr.sh
+    # Set file and directory permissions
+    #chmod 666 sites/default/sqlconf.php why?
```
bradymiller: After the install, agree this is treated just like all the other normal scripts(just needed to be writable during the install as you did above).

docker-images/openemr/run_openemr.sh
```diff
@@ -1,9 +1,53 @@
...
+    chmod -R 600 .
+    php auto_configure.php -f \
+        iuser=$OE_USER \
+        iuserpass=$OE_PASS \
+        login=$MYSQL_USER \
+        rootpass=$MYSQL_ROOT_PASS \
+        pass=$MYSQL_PASS \
+        server=$MYSQL_HOST \
+
+    #do I need to activate mod_rewrite?
```
bradymiller: As I recall, you dealt with this in the apache conf script (note you can testing by ensuring just don't get error screen when go to the Modules menu item in OpenEMR).

docker-images/openemr/run_openemr.sh
```diff
@@ -1,9 +1,53 @@
...
+
+    echo "Removing remaining setup scripts"
+    #remove all setup scripts
+    rm acl_setup.php
+    rm acl_upgrade.php
+    rm setup.php
+    rm sql_upgrade.php
+    rm ippf_upgrade.php
+    rm gacl/setup.php
+    rm auto_configure.php
```
bradymiller: removing these is definitive. i like it :)

docker-images/openemr/run_openemr.sh
TheToolbox: Oh whoops. That line was actually my personal note and one I meant to remove. I'll take that out when I clean this up in the morning.

docker-images/openemr/run_openemr.sh
TheToolbox: I'll go ahead and remove that comment then!

docker-images/openemr/run_openemr.sh
TheToolbox: I'll take the comment out then. I don't even want to know why zend needs write access to its config file after it's set up...

docker-images/openemr/run_openemr.sh
TheToolbox: Yeah it took me some time to figure out it needed to be 700 and not 600. I feel like I need to learn a bit more deeply why directories behave the way they do w.r.t. the execute bit.

docker-images/openemr/Dockerfile
TheToolbox: I came across an issue in docker's github repo that suggested that VOLUME was relative to WORKDIR  but honestly it can't hurt to make it absolute so I might as well. I'll do that in the cleanup commit in the morning!

PR #16, opened 2017-10-02
More predictable config startup behavior

TheToolbox:
Now it pitches a fit and loops until it can connect to mysql when doing autoconfig


TheToolbox:
So odd... The build for 0759c09 failed at first on `apk add`... I guess it was a weird fluke?

PR #19, opened 2017-10-06
Appliance Lockdown

jesdynf:
* adds appliance password-change support
 * launch.sh now accepts command-line parameters


docker-images/openemr/Dockerfile
```diff
@@ -17,14 +17,15 @@ ENV APACHE_LOG_DIR=/var/log/apache2
 COPY php.ini /etc/php7/php.ini
 COPY openemr.conf /etc/apache2/conf.d/
 #add runner and auto_configure and prevent auto_configure from being run w/o being enabled
-COPY run_openemr.sh .
-COPY auto_configure.php .
-RUN chmod 700 run_openemr.sh \
-    && chmod 000 auto_configure.php \
+COPY run_openemr.sh auto_configure.php utilities/unlock_admin.php /var/www/localhost/htdocs/openemr/
```
bradymiller: Are you copying these scripts into the webroot at /var/www/localhost/htdocs/openemr/ ?
(if so, that doesn't seem very safe; can they work if placed outside the web path)

jesdynf: I thought `sites` below it was the webroot; I must've been mistaken. Isn't this where auto_configure.php goes? I don't know if it'll run when outside the webroot or not -- is there somewhere else you'd like it?

bradymiller: Also it seems like the unlock_admin.php is just basically specific to lightsail. Does it make sense to include this in the main docker where it will mostly not be utilized?

docker-images/openemr/Dockerfile
```diff
@@ -17,14 +17,15 @@ ENV APACHE_LOG_DIR=/var/log/apache2
...
+COPY utilities/unlock_admin.sh /root
```
bradymiller: same question as above (since this is specific to lightsail)

docker-images/openemr/Dockerfile
jesdynf: I considered that, but decided that we're probably going to come up with more random little nubs, and remember that it has to be inside the container to work anyways, so if I don't put it in the Docker now I'll have to inject it from the Lightsail host. It just seemed cleaner to put it here.

jesdynf: Also, it's not quite that it's specific to Lightsail -- this isn't code for Lightsail at all, it's for the AWS MArketplace appliance. It's just being built from the Lightsail path.

bradymiller: Looked into this more.
auto_configure gets removed after an autoinstall, but is left on a manual install. We need to also remove it for a manual install (otherwise, it is a security risk). and would place  unlock_admin.php in /root/ and run it from there (it will work anywhere).

bradymiller: And then if the unlock_admin.php is in /root/ and permissions are locked, then we'd be safe to make it ubiquitous for all dockers.

docker-images/openemr/Dockerfile
bradymiller: disregard this post since can just place the unlock_admin.php in this /root/ dir and lock the permissions down, so then safe to include this hook in all dockers.

docker-images/openemr/Dockerfile
jesdynf: Done.

bradymiller:
also posted a quick minor security improvement here:
https://github.com/openemr/openemr-devops/pull/20

bradymiller:
This looks good to me

stacks/single-server/launch.sh
```diff
@@ -1,16 +1,48 @@
 #!/bin/bash
 
+# OpenEMR Cloud Express launcher
+# usage: launch.sh -s 2 -b wip-feature -d
+#        -s: amount of swap to allocate, in gigabytes
+#        -b: repo branch to load instead of master
+#        -d: start in developer mode, force local dockers and open ports
```
TheToolbox: This flag might require more description. I'm not sure what 'force local dockers' means and you may want to specify if you're opening container or host ports.

docker-images/openemr/Dockerfile
```diff
@@ -17,14 +17,14 @@ ENV APACHE_LOG_DIR=/var/log/apache2
 COPY php.ini /etc/php7/php.ini
 COPY openemr.conf /etc/apache2/conf.d/
 #add runner and auto_configure and prevent auto_configure from being run w/o being enabled
-COPY run_openemr.sh .
-COPY auto_configure.php .
-RUN chmod 700 run_openemr.sh \
-    && chmod 000 auto_configure.php \
+COPY run_openemr.sh auto_configure.php /var/www/localhost/htdocs/openemr/
+COPY utilities/unlock_admin.php utilities/unlock_admin.sh /root/
```
TheToolbox: not sure that it needs to be in a utilities folder all by its lonesome, since all of the other files there are either config files or utilities. An argument could be made for moving all of them into folders with those names though.

stacks/single-server/appliance-unlock.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
TheToolbox: Very much might be wrong here, but using the word 'appliance' for the single-server setup stuff seems like it might be confusing to end-users. I realize it probably fits the criteria for being called an appliance, but perhaps another name might be more appropriate? ('instance-unlock' or just 'set-admin-password' or something to that effect?)

docker-images/openemr/Dockerfile
```diff
@@ -17,14 +17,14 @@ ENV APACHE_LOG_DIR=/var/log/apache2
...
+    && chmod 000 auto_configure.php /root/unlock_admin.php \
     && ln -s /usr/bin/php7 /usr/bin/php
 #fix issue with apache2 dying prematurely
 RUN mkdir -p /run/apache2
 #go
 CMD [ "./run_openemr.sh" ]
 
-EXPOSE 80
\ No newline at end of file
+EXPOSE 80
```
TheToolbox: Good catch here. I think I may have read somewhere about Dockerfiles technically requiring a blank final line to even be valid.

stacks/single-server/launch.sh
jesdynf: It's declining to use the openemr at the hub, and forcing the use of the one in the current branch of the repo instead, the one that just got checked out. It's also forcing the MySQL container to open 3306 to the host; usually this isn't done.

docker-images/openemr/Dockerfile
jesdynf: I looked at the "fragment of useful PHP" and I had a hunch we'd pick up more of them -- and I wanted them out of the top level so that it'd be a more pleasant read, with fewer moving parts disclosed to a newcomer on first glance.

stacks/single-server/appliance-unlock.sh
jesdynf: Could, but it's targeted at AWS Marketplace compliance, so it's only ever going to be used when it's serving a role *as* an appliance, which is why I named it like that. Don't feel strongly about it, though.

jesdynf: In other words, the only time it'll ever be used is in the post-deployment Marketplace notes 
when a user is explicitly told to use it.

stacks/single-server/launch.sh
TheToolbox: Oh sorry. I meant that it might be useful to be more explicit about that in the help page there

docker-images/openemr/Dockerfile
TheToolbox: Should we not then also move auto_configure.php in there as well?

stacks/single-server/appliance-unlock.sh
TheToolbox: Perhaps having it in the main docker repo is the wrong place for it then, if it's only ever expected to be used in the docker appliance. It might then make more sense to inject it in as needed.

docker-images/openemr/Dockerfile
jesdynf: Not in my view, because it's going to be used most of the time -- I looked at it and thought 'load-bearing', it should be presented to the reader as a thing of value at the top level. If primary install involved a lot of php fragments I'd've considered it differently.

stacks/single-server/appliance-unlock.sh
jesdynf: Maybe, but I'm going to replace it with a GENERALIZED password recovery script as soon as I can manage it, so it'll go in the same place.

docker-images/openemr/Dockerfile
TheToolbox: As good a reason as any. Everything else looks good to me.

bradymiller: @TheToolbox ,
I thought above needing to move auto_configure also, but it was using relative paths, so figured that would then require more testing :)
Also, felt safe since it is always removed.

bradymiller: Take that back. It doesn't use relative paths. It could be moved, but again, it is always deleted.

PR #22, opened 2017-10-06
Ensuring permissions settings are maintained if container is recreated

### LINKED ISSUE #21: `rm auto_configure.php` causes crash after auto-config
Looks like by adding the second `rm auto_configure.php` [here](https://github.com/openemr/openemr-devops/blob/c6abcdab46eab77bcd1c53fa7aed2a186a07f857/docker-images/openemr/run_openemr.sh#L83), it's trying to remove the file twice. Removing it the first time seems likely to be sufficient, although I have to test the behavior of stopping and restarting a container to better understand which things will be persisted. Additionally, we may need to take some steps to properly persist the permissions that are done during autoconfig. Presently, they may be lost when a container is killed, but its site volume is preserved, and a new container is started in its place. The logical solution seems to be to move them into a separate 'if' block, checking only if `$CONFIG == 1` and then re-applying the permissions forcefully. That would honestly take care of both issues I think.

TheToolbox:
This also fixes #21

### DIFF
--- a/docker-images/openemr/run_openemr.sh
+++ b/docker-images/openemr/run_openemr.sh
@@ -32,34 +32,12 @@ auto_setup() {
     chmod -R 600 .
     php auto_configure.php -f ${CONFIGURATION} || return 1
 
-    echo "OpenEMR configured. Setting user 'www' as owner of openemr/ and setting file/dir permissions to 400/500"
-    #set all directories to 500
-    find . -type d -print0 | xargs -0 chmod 500
-    #set all file access to 400
-    find . -type f -print0 | xargs -0 chmod 400
-
-    echo "Default file permissions and ownership set, allowing writing to specific directories"
-    chmod 700 run_openemr.sh
-    # Set file and directory permissions
-    chmod 600 interface/modules/zend_modules/config/application.config.php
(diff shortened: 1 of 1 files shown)


docker-images/openemr/run_openemr.sh
```diff
@@ -32,7 +32,33 @@ auto_setup() {
...
+        echo "Couldn't set up. Any of these reasons could be what's wrong:"
+        echo " - You didn't spin up a MySQL container or connect your OpenEMR container to a mysql instance"
+        echo " - MySQL is still starting up and wasn't ready for connection yet"
+        echo " - The Mysql credentials were incorrect"
+        sleep 1;
+    done
+    echo "Setup Complete!"
+fi
+
+if [ "$CONFIG" == "1" ]; then
```
bradymiller: Doesn't all the stuff also need to happen after the autoinstall?
(I think above if will skip that scenario)

TheToolbox: I double checked. It does correctly happen after autoinstall (since autoinstall now sets CONFIG again when it completes)

bradymiller: oh, that makes sense (forgot that a function doesn't limit scope of a variable in bash). looks good to go then.

PR #23, opened 2017-10-07
Versioning reorganization

TheToolbox:
Moved the openemr:5.0.0 image into a 5.0.0 folder, and adjusted the docker-compose-dev file in the single-stack (which was the only file I noticed that cared about that change). Hopefully it's all good?


TheToolbox:
Oh right. I forgot this would break CI. Sending a fix...

stacks/single-server/docker-compose.dev.yml
```diff
@@ -13,7 +13,7 @@ services:
       MYSQL_ROOT_PASSWORD: root
   openemr:
     restart: always
-    build: ../../docker-images/openemr/
+    build: ../../docker-images/openemr/5.0.0
```
jesdynf: You removed the trailing slash, but I can't tell if it's needed or not.

TheToolbox: I'll run a build to make sure it's not

TheToolbox: Confirmed it's all good. Now if I would just stop making foolish mistakes in the ci config...

Issue #26, opened 2017-10-08
Docker support for Let's Encrypt


jesdynf:
We need to support getting proper certificates on the instance after installation.

PR #38, opened 2017-10-19
Adding SSL

### LINKED ISSUE #25: Self-Signed SSL Support
The OpenEMR docker should generate and install a self-signed SSL certificate, and expose 443.

TheToolbox:
This updates the openemr dockerfile, exposing TLS https on port 443.
When provided with `$DOMAIN` and `$EMAIL` it will pull down a certificate from letsencrypt.
Otherwise, it generates a self-signed certificate. 

The LetsEncrypt functionality still needs significant testing, but the self-signed cert generation works as expected.

We also may want to look into adding some more restrictive SSL settings to the apache config (for example only allowing TLS1.2), but that may land in a further PR.

This closes #25 and #26


bradymiller:
Very nice. If self-signing is testing well, then lets gets this in. The letsencrypt part is also a really nice touch!

TheToolbox:
Build is failing on xtradb. Looks like there's a package that is no longer available?
Also sorry for the 7 commits. I made a bit of a mess in my fork, so I recommend just looking at the overall changes.

TheToolbox:
and because I know this was important, the self-signed cert is generated at runtime. As soon as @jesdynf gives it a thumbs-up I'll bring it on in.

docker-images/openemr/5.0.0/run_openemr.sh
```diff
@@ -89,5 +118,7 @@ fi
 # ensure the auto_configure.php script has been removed
 rm -f auto_configure.php
 
+echo "Starting cron daemon!"
+crond
```
jesdynf: What's going on here? Is certbot using cron in a non-obvious way?

docker-images/openemr/5.0.0/Dockerfile
```diff
@@ -27,4 +37,4 @@ RUN mkdir -p /run/apache2
 #go
 CMD [ "./run_openemr.sh" ]
 
-EXPOSE 80
+EXPOSE 80 443
```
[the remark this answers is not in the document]
TheToolbox: ah right. I always absentmindedly kill those newlines. docker didn't complain but i'll throw one in for good measure

docker-images/openemr/5.0.0/run_openemr.sh
TheToolbox: it's using it in the obvious way. This is just starting the cron daemon

jesdynf: What's using the cron daemon, though? We're not maintaining any crontabs ourselves.

jesdynf:
What are some example command lines? How to launch it from the Docker-compose hasn't changed, but if I want to, say, recycle SSL certs from the host command line, how do I do that?

docker-images/openemr/5.0.0/run_openemr.sh
TheToolbox: oh whoops. I forgot to actually pipe the echo on line 20 anywhere. It runs the renewal script every day to check if certs need to be renewed. Fixing shortly...

TheToolbox:
By default, the image throws the certs into a volume, so they will persist across container starts.

I've just been using `sudo docker run -p 8080:80 -p 8081:443 openemr` to test it. Works like a charm (for self-signed at least)

PR #54, opened 2017-10-24
switch to Community AMIs

danielehrlich:
can't edit file directly so creating pull request


jesdynf:
Sorry for the confusion -- when we release to the Marketplace, it'll be right.

danielehrlich:
ahh okay got it, thanks Asher

PR #56, opened 2017-10-25
Adding 5.0.1 docker image

TheToolbox:
Notes:
 - migrates to alpine:3.6 which has a simpler process for apache/php install
 - all env inputs are maintained, travis should already check the build status
 - recommended tags: 'next', 'dev'


jesdynf:
So /stacks should still be using openemr:5.0.0, right?

TheToolbox:
Yep, until OEMR 5.0.1 is released.

bradymiller:
@jesdynf ,
Stacks should only use 5.0.0. 5.0.1 won't be ready for official release for probably another 2 months or so.

@TheToolbox ,
Vert good idea to get this out there. I'd be happy to build this (will use the `dev` tag) and push it to docker hub.
As an aside, what are your thoughts on a 'flex' openemr docker that allows the 2 following env variable:
repository (ie. 'https://github.com/openemr/openemr.git' for example)
branch (ie. 'master' for example)
(On new startup, it would then grab the repo/branch and use this for install; so could basically test any git code out there; would be awesome nifty trick for developers/testers)

bradymiller:
Once this is merged in, I'll gladly build this and push it to docker hub. :)

PR #59, opened 2017-10-28
added tar and mysql-client to dockers

bradymiller:


bradymiller:
whoohooo, testing well and will likely get the new docker up by the am.

PR #65, opened 2017-11-03
Structural reorganization to match offering names

jesdynf:
This is riddled with todos, but I need it out of the branch anyways so I can rebuild the AMIs, and then go back and fix all the docs.


bradymiller:
very nice work!

jesdynf:
We're not hosting images, just build files. They're images after they're assembled. Also it's shorter -- no need for a compound word with a dash; the meaning is still 100% clear, herein live Docker-things.

PR #105, opened 2018-01-21
Update php eb version for full stack

MatthewVita:


MatthewVita:
Asher, I had to update Troposphere. It was way to old to run. Was saying it didn't support the `Encrypted` keyword arg for EFS ha

PR #108, opened 2018-03-05
quick fix for "try it" example for flex docker

bradymiller:


bradymiller:
@TheToolbox , Just a quick fix for the Try This link in the flex README

PR #109, opened 2018-03-10
removing php-mcrypt and using php-openssl instead

bradymiller:
removing php-mcrypt and using php-openssl instead, since mcrypt not supported in PHP 7.2
(testing well and already updated the dockers with this)


bradymiller:
@TheToolbox and @jesdynf ,
This is just a quick change in encryption php packages since new versions of php are dropping mycrypt (openemr codebase has just recently been changed to openssl). Already have updated the dockers (not just for 5.0.1-dev and flex; 5.0.0 still will keep using mycrypt).

PR #113, opened 2018-04-08
add mysql volume

aptalca:
mysql container should be ephemeral. Without a separate volume for db data, a recreation of the container deletes the entire database (or even a docker-compose down)

### DIFF
--- a/docker/openemr/5.0.0/README.md
+++ b/docker/openemr/5.0.0/README.md
@@ -25,6 +25,8 @@ services:
     restart: always
     image: mysql
     command: ['mysqld','--character-set-server=utf8']
+    volumes:
+    - databasevolume:/var/lib/mysql
     environment:
       MYSQL_ROOT_PASSWORD: root
   openemr:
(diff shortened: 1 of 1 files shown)


jesdynf:
It wasn't my intent to allow rebuilding the containers, but I think you're right. Give me a little time to review -- thanks!

MatthewVita:
@aptalca Ah, we missed this. Your PR looks correct to me but I'll leave the rest to @jesdynf for review.

MatthewVita:
Also, welcome to OpenEMR!

bradymiller:
Looks good to me also. If we do this, then will also make sense to also do for the 5.0.1, 5.0.2, and flex dockers.

MatthewVita:
Great stuff!

MatthewVita:
(BTW, I'm not adding this to the local development solution - that whole thing is meant to be ephemeral)

bradymiller:
@jesdynf , do you want this on the other docker README's. If so, I can add this quickly(and will update the try this scripts also).

PR #117, opened 2018-04-14
replace links with depends_on

aptalca:
`links` is deprecated in docker: https://docs.docker.com/compose/compose-file/#links
Docker-compose v3 will already create a user defined bridge network and put both services in there so they will be linked.

`depends_on` will declare mysql as a dependency for openemr and docker-compose will make sure it is started first in any scenario (restart, update, recreation, etc.) and is a substitute for `links` for dependency control

One thing to keep in mind is that this will not share environment variables between containers. Not sure if that is something these images utilize.

Not a crucial change but if you approve, I can go ahead and submit PRs for this change as well as adding the mysql volume to 5.0.1, 5.0.2 and flex

### DIFF
--- a/docker/openemr/5.0.0/README.md
+++ b/docker/openemr/5.0.0/README.md
@@ -45,7 +45,7 @@ services:
       MYSQL_PASS: openemr
       OE_USER: admin
       OE_PASS: pass
-    links:
+    depends_on:
     - mysql
 volumes:
   logvolume01: {}


jesdynf:
I really appreciate you taking the time to look this over! Docker isn't my strongest tool, and I did miss these completely. That said, please don't worry about the PR for the MySQL volume in the containers; I've already patched that into the compose files, and since we don't /always/ want the volume to persist (like in the developer compose variants) I'd prefer to keep it undeclared in the container.

bradymiller:
hi @jesdynf ,
I thought we needed to go back to v2 for docker-compose or did I misunderstand things?

jesdynf:
We do -- but I checked, and v2 /also/ doesn't like `links`, it calls it 'legacy'.

bradymiller:
@jesdynf , @TheToolbox , @aptalca ,
Looks like play with docker site (where the try this links run on) do not work with v2 (since docker swarm it appears), so not gonna touch the version that is currently being used. On reading docs, it appears we should be supporting v3 anyways: https://docs.docker.com/compose/compose-file/compose-versioning/
(looks like mem_limit has a new mechanism in it, if needed)

TheToolbox:
The issue with mem_limit in v3 is that it is not imposed when using docker-compose, and only runs with `docker stack`. I'll go ahead and see if maybe there's a more appropriate toolchain we should be supporting, but I think it's pretty reasonable to hold onto v2, especially in light of them saying they intend to continue supporting it indefinitely. And great stuff @aptalca!

PR #122, opened 2018-04-19
update permissions and remove scripts only once

bradymiller:
hi @jesdynf and @TheToolbox ,
During the board meeting yesterday, Jit discussed some issues with the dockers. One issue was that everytime they restarted the docker, all the permissions (and script removals), were changed (rather than just doing once after an install/configuration). It seems reasonable to only do this once (and not make users have to modify these everytime they restart the docker. Here's my proposal to do this(I still haven't tested this, so don't bring this into codebase yet). Hoping to get your thoughts?
(@sjpadgett , since you asked a related question regarding this, also mentioning you on this :) )
thanks, -brady

### DIFF
--- a/docker/openemr/5.0.0/autoconfig.sh
+++ b/docker/openemr/5.0.0/autoconfig.sh
@@ -14,15 +14,15 @@ if ! [ -f /etc/ssl/private/selfsigned.key.pem ]; then
     -keyout /etc/ssl/private/selfsigned.key.pem \
     -out /etc/ssl/certs/selfsigned.cert.pem \
     -days 365 -nodes \
-    -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=localhost" 
+    -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=localhost"
 fi
 rm -f /etc/ssl/certs/webserver.cert.pem
 rm -f /etc/ssl/private/webserver.key.pem
-ln -s /etc/ssl/certs/selfsigned.cert.pem /etc/ssl/certs/webserver.cert.pem 
-ln -s /etc/ssl/private/selfsigned.key.pem /etc/ssl/private/webserver.key.pem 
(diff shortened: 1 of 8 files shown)


bradymiller:
btw, so see things much more clearly, use the ?w=1 flag in the link:
https://github.com/openemr/openemr-devops/pull/122/files?w=1

jesdynf:
I can't test right now, but the change seems reasonable. The failed Travis test is a network hiccup (I think) in the MySQL docker build.

bradymiller:
Just requested a travis rebuild.

I'll plan to test this extensively, especially the MANUAL_SETUP setting to ensure no weirdness with that.

jesdynf:
If the rebuild isn't green, I'll investigate it this evening -- note that this docker build failure won't affect anybody in production.

jesdynf:
Well, they did release a new 5.7 patch on April 18th (5.7.22-1debian9), so I guess I can refer to that one by name, but that shouldn't mean build *fails*, that's still a legal version.

Ugh, whatever, I'll patch it tonight.

TheToolbox:
Definitely looks reasonable to me, although I'm curious why this was an issue for Jit

bradymiller:
practical scenario would be:
There is another directory in sites/default/ or even sites/default itself that the user wished to make writable. The problem is the user needs to then set these every time the docker restarts.

bradymiller:
This is testing well.
I also brought in the fix from this issue here(the second commit), which is also testing well:
https://github.com/openemr/openemr-devops/issues/124

bradymiller:
Updated the dockers (since getting very close to release, expedited this to ensure gets some testing) with this.

jesdynf:
Merge master back into this PR to clear the failing test.

bradymiller:
just did a rebase(master)/push, which should clear things up

bradymiller:
testing well and need this for release, so bringing it in.

PR #125, opened 2018-04-22
Fix mysql upgrade

jesdynf:
pick up MySQL repo upgrade


bradymiller:
Just checking if this docker needs to be rebuilt and pushed to docker hub?

jesdynf:
It does, please.

PR #129, opened 2018-04-26
change the example docker-compose scripts from mysql to mariadb

bradymiller:
Turns out that if we do mysql image (that defaults to latest tag), it is now using mysql 8.x which causes issues:
1. main issue being mysql 5.6/5.7 client is not compatible with mysql 8 server (get some login authentication not supported issues). so can't really do this until most operating systems are using mysql 8
2. also brought the demo farm to its knees resource wise

2 options are using mysql:5.7 or going to mariadb docker. Ended up going with mariadb, for now, which works very well (from their latest tag docker which uses a version that maintains compatibiltiy with mysql 5.6/5.7).

This is testing well.

### DIFF
--- a/docker/openemr/5.0.0/README.md
+++ b/docker/openemr/5.0.0/README.md
@@ -23,7 +23,7 @@ version: '3.1'
 services:
   mysql:
     restart: always
-    image: mysql
+    image: mariadb
     command: ['mysqld','--character-set-server=utf8']
     volumes:
     - databasevolume:/var/lib/mysql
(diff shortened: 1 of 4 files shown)


jesdynf:
Sounds good. Note, though, that Lightsail (and all the derivatives) are
using MySQL 5.7.22-1debian8, and that I'd have to do research before I was
prepared to switch them to MariaDB, because I've got the XtraBackup
hot-backup tech in there and I'd need to verify it works alike or find a
replacement. mysql:5.7 may be a better long-term choice.

-- 
Asher Densmore-Lynn <jesdynf@gmail.com>

TheToolbox:
Seems reasonable to me. For reference, the [MariaDB incompatabilities w/ MySQL](https://mariadb.com/kb/en/library/incompatibilities-and-feature-differences-between-mariadb-102-and-mysql-57/#incompatibilities). It's a bit of an oversight on my part that we weren't specifying a mysql version in the first place, which is definitely good practice. Do we know of any installations using MariaDB?

bradymiller:
Alpine actually uses mariadb and lots of folks use mariadb rather than mysql(the xampp package also uses mariadb). The alpine is interesting since the package is called mysql, but the version is uses means it is actually mariadb:
https://pkgs.alpinelinux.org/packages?name=mysql&branch=v3.7

I definitely agree we should not change any of the cloud stuff (ie. keep using mysql with the xtrabackup). The real goal is for OpenEMR to always support both mysql/mariadb, so will be good to use both (mysql in the cloud stuff and mariadb in the docker-compose.yml examples (note these are really just examples, anyways and starting points for folks).

The above link is interesting since mariadb does not support json very well and simply stores them in a longtext, which isn't really a big deal anyways (it is pertinent since at some point we will be migrating the json menus to database and I was planning to store the json as a longtext anyways).

bradymiller:
Hold off on bringing this in for now. Still contemplating mysql:5.7 :)
(will have more details later)

TheToolbox:
The one suggestion I might have is to pin the mariadb version to whatever latest currently is, and we will manually push updated versions after we test they work, although that wouldn't be a concern for dev builds.

bradymiller:
@TheToolbox ,
Agreed. I'll add this (and will also do the same for the demo farm).
-brady

bradymiller:
@TheToolbox ,
I added mariadb version tag for docker and working well on all testing.
-brady

bradymiller:
Gonna bring this in before I forget about it :)

PR #132, opened 2018-05-07
support an empty flex docker for some more development flexibility

bradymiller:
@TheToolbox , @jesdynf , @d3sandoval 
These are some mods to flex to make it more flexible to use in development modes. Added a EMPTY environment which basically means is does not grab openemr since the user instead can set it to a local volume. See the example docker-compose.yml to see how i am using it in my development environment:
https://gist.github.com/bradymiller/92ab6b8c401baa129130dbc7b21f98b6


bradymiller:
Still working out some issues on this (works well in EMPTY mode, but I am sorting out permission issues in non-EMPTY mode).

bradymiller:
ok, finally got it working in both modes and is ready to get into mainline.
1. Bringing in any git repo from internet
2. Making it empty and instead using local copy of openemr

Have uploaded this new flex docker to docker hub

Here's the docker-compose.yml that I am using for development environment now:
https://gist.github.com/bradymiller/92ab6b8c401baa129130dbc7b21f98b6
(note it includes instructions on how to setup)

bradymiller:
@TheToolbox and @jesdynf ,
Ok, I've added something here both a bit drastic and cool. In order to support simultaneous development/testing with PHP 7.1 and 7.2 I created a flex-edge docker which uses Alpine Edge (and is using PHP 7.2). This then allows using both on the same place as done here:
https://gist.github.com/bradymiller/92ab6b8c401baa129130dbc7b21f98b6

This is testing well and is pretty neat. Can build upon it more with log volumes etc.

bradymiller:
Here's the companion PR in the main codebase related to this:
https://github.com/openemr/openemr/pull/1613

TheToolbox:
Looks great to me! Feel free to bring it in whenever, or I'll do it in 24 hours or so.

PR #135, opened 2018-05-09
some more things in flex series dockers so more developer/tester friendly

bradymiller:
ok, i swear, I think this will be the last flex docker stuff for awhile


bradymiller:
dockers updated and I hope this is the last work I do on flex series for awhile (or at least the next day :) )

PR #138, opened 2018-05-30
Fix for updates to core alpine packages

TheToolbox:
The `apache2-ssl` package was throwing a fit when trying to install, due to its post-install hook throwing 127. References online seemed to suggest it was due to a libressl change, and could be fixed by first doing an `apk upgrade` before attempting installation. This seems like reasonable behavior to add to all images (in case upgrades to musl/libressl/apk-tools/etc need to be picked up in the future). If this passes all builds, I think it should be good, but there's no rush to push a new image to docker hub.


TheToolbox:
Yeah, foolish oversight on my part, but I suppose no harm, no foul.

PR #139, opened 2018-05-30
Update README to denote key pair creation

JXPrime:
If you don't have a key pair in the region you select, the installation will let you proceed without one and the Stack creation will fail. Added language to call out this fact.


jesdynf:
Nice catch, thanks! It's really frustrating that I can't require a valid keypair as part of the parameter validation in CloudFormation. (Or if I can, it's never been apparent to me /how/.)

PR #141, opened 2018-06-01
Fixing email detection

TheToolbox:
Fixes inverted warning for no $EMAIL when $DOMAIN is set, and fixes a bug causing LE to fail in that same scenario.

### DIFF
--- a/docker/openemr/5.0.0/Dockerfile
+++ b/docker/openemr/5.0.0/Dockerfile
@@ -2,6 +2,7 @@ FROM alpine:3.5
 
 #Install dependencies
 #php7-tokenizer will be needed when migrating to 7.1
+RUN apk --no-cache upgrade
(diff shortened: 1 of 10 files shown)


TheToolbox:
Okay a bit confused (probably just tired). I accidentally included the "apk --update" commit from the apache-ssl fix, whcih I figured was no big deal, given that I was going to squash them anyway. Why is github showing those as changed files with respect to master? I just double-checked that master has those lines in it. Am I missing something?

jesdynf:
I saw that and I don't get it either, but the commit seems fine. Let's wait on Travis, though.

TheToolbox:
Travis is good, bringing it in (with hopefully no issues...)

bradymiller:
@TheToolbox , I just updated all the dockers and pushed them to docker hub :)

PR #143, opened 2018-06-16
Update docker shell access

JXPrime:
sudo is needed in both instances of the docker command (I cut and paste this all the time lol)

### DIFF
--- a/packages/standard/README.md
+++ b/packages/standard/README.md
@@ -6,7 +6,7 @@
 
 ## Administration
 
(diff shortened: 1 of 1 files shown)


JXPrime:
That's a bit beyond mu linux-fu. the other option that I can think of is to
do sudo -i and then the rest of the command?

jesdynf:
I think I got it--

`sudo -i -- sh -c 'docker exec -it $(docker ps | grep _openemr | cut -f 1 -d " ") /bin/sh'`

Can you update to pick that up?

JXPrime:
Oh brilliant! I have so much to learn still. Thanks!

jesdynf:
Hey, thanks for the improvement!

PR #144, opened 2018-06-23
changes to allow sharing website volume with nginx and php-fpm dockers

bradymiller:
@TheToolbox and @jesdynf ,

I was able to solve the problem of getting shared volumes across openemr, nginx, and php-fpm dockers to work. There were 2 issues:

1. The first was that all the owner uid's need to be the same. After some research, it looks like a option is to set the uid to 1000 for applicable users (in openemr docker is apache user, in nginx docker is nginx user, in php-fpm docker is www-data user); here are corresponding changes in the nginx/php-fpm dockers: https://github.com/openemr/openemr/commit/df4ab4945804d73c87f50978e6030404840bde04

2. Shared volumes do not work when they overlap. So the volume setting in the dockerfile for /var/www/localhost/htdocs/openemr/sites was basically not allowing me to create a volume for /var/www/localhost/htdocs/openemr/ . Super strange it seemed, but I confirmed this by showing that the path for the shared volume (via inspect) was not getting updated. This forced me to need to remove the /var/www/localhost/htdocs/openemr/sites from the dockerfile, which seems reasonable, since this is not needed and really does not do anything (can instead just set up the shared volumes in the docker-compose.yml script)

Here is the docker-compose.yml that is being used to glue together the openemr/mysql/nginx/php-fpm dockers just so you can see what I am doing:
https://gist.github.com/bradymiller/fb8f0fbf9fff74b7666a217f0e9c28a1

Thoughts?

### DIFF
--- a/docker/openemr/5.0.2/Dockerfile
+++ b/docker/openemr/5.0.2/Dockerfile
@@ -7,7 +7,8 @@ RUN apk add --no-cache \
     php7-json php7-pdo php7-pdo_mysql php7-curl php7-ldap php7-openssl php7-iconv \
     php7-xml php7-xsl php7-gd php7-zip php7-soap php7-mbstring php7-zlib \
     php7-mysqli php7-sockets php7-xmlreader perl mysql-client tar curl imagemagick-dev \
-    python openssl git libffi-dev py-pip python-dev build-base openssl-dev dcron shadow
+    python openssl git libffi-dev py-pip python-dev build-base openssl-dev dcron \
+    shadow
 # Needed to ensure permissions work across shared volumes with openemr, nginx, and php-fpm dockers
 RUN usermod -u 1000 apache
 #clone openemr
--- a/docker/openemr/5.0.2/Dockerfile
+++ b/docker/openemr/5.0.2/Dockerfile
@@ -7,7 +7,9 @@ RUN apk add --no-cache \
     php7-json php7-pdo php7-pdo_mysql php7-curl php7-ldap php7-openssl php7-iconv \
     php7-xml php7-xsl php7-gd php7-zip php7-soap php7-mbstring php7-zlib \
     php7-mysqli php7-sockets php7-xmlreader perl mysql-client tar curl imagemagick-dev \
-    python openssl git libffi-dev py-pip python-dev build-base openssl-dev dcron
+    python openssl git libffi-dev py-pip python-dev build-base openssl-dev dcron shadow
+# Needed to ensure permissions work across shared volumes with openemr, nginx, and php-fpm dockers
+RUN usermod -u 1000 apache
 #clone openemr
 RUN git clone https://github.com/openemr/openemr.git --depth 1 \
     && rm -rf openemr/.git \
@@ -20,7 +22,7 @@ RUN git clone https://github.com/openemr/openemr.git --depth 1 \
     && mkdir -p /etc/ssl/certs /etc/ssl/private \
     && apk del --no-cache git build-base libffi-dev python-dev
 WORKDIR /var/www/localhost/htdocs/openemr
-VOLUME [ "/var/www/localhost/htdocs/openemr/sites", "/etc/letsencrypt/", "/etc/ssl" ]
+VOLUME [ "/etc/letsencrypt/", "/etc/ssl" ]
 #configure apache & php properly
 ENV APACHE_LOG_DIR=/var/log/apache2
 COPY php.ini /etc/php7/php.ini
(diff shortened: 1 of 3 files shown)


bradymiller:
btw,
Only plan to make these changes to 5.0.2 and flex dockers for now (won't mess with the 5.0.1 production docker).
-brady

jesdynf:
You're changing the user of the Apache user, but won't that leave a million orphaned files anywhere Apache used to own things, under the old numerical uid? You don't need to chase those down at all?

bradymiller:
Actually, I may be able to get away with setting uid to 100 in the nginx/php-fpm dockers. Let me try that then the apache change to 1000 won't be needed.

bradymiller:
dang, no go. that uid is already being used in the nginx docker. Not seeing any errors with the apache uid to 1000, but will do some more research/analysis on it.

jesdynf:
I don't know what Apache's uid is pre-swap, but you might do a find and hunt down all the files that bear it -- and if you're already doing that, maybe chown the files and directories to 1000.

bradymiller:
pre-swap it is 100 and the following command finds nothing:
``````
find / -user 100
``````

bradymiller:
whereas:
``````
find / -user 1000
``````
finds all the stuff in /var/www/html/openemr dir and also:
``````
/proc/112
/proc/112/task
/proc/112/task/112
/proc/112/task/112/net
/proc/112/task/112/attr
/proc/112/task/112/attr/selinux
/proc/112/task/112/attr/smack
/proc/112/task/112/attr/apparmor
/proc/112/net
/proc/112/attr
/proc/112/attr/selinux
/proc/112/attr/smack
/proc/112/attr/apparmor
/proc/113
/proc/113/task
/proc/113/task/113
/proc/113/task/113/net
/proc/113/task/113/attr
/proc/113/task/113/attr/selinux
/proc/113/task/113/attr/smack
/proc/113/task/113/attr/apparmor
/proc/113/net
/proc/113/attr
/proc/113/attr/selinux
/proc/113/attr/smack
/proc/113/attr/apparmor
/proc/114
/proc/114/task
/proc/114/task/114
/proc/114/task/114/net
/proc/114/task/114/attr
/proc/114/task/114/attr/selinux
/proc/114/task/114/attr/smack
/proc/114/task/114/attr/apparmor
/proc/114/net
/proc/114/attr
/proc/114/attr/selinux
/proc/114/attr/smack
/proc/114/attr/apparmor
/proc/115
/proc/115/task
/proc/115/task/115
/proc/115/task/115/net
/proc/115/task/115/attr
/proc/115/task/115/attr/selinux
/proc/115/task/115/attr/smack
/proc/115/task/115/attr/apparmor
/proc/115/net
/proc/115/attr
/proc/115/attr/selinux
/proc/115/attr/smack
/proc/115/attr/apparmor
/proc/116
/proc/116/task
/proc/116/task/116
/proc/116/task/116/net
/proc/116/task/116/attr
/proc/116/task/116/attr/selinux
/proc/116/task/116/attr/smack
/proc/116/task/116/attr/apparmor
/proc/116/net
/proc/116/attr
/proc/116/attr/selinux
/proc/116/attr/smack
/proc/116/attr/apparmor
``````

bradymiller:
This likely works so nicely since the uid of apache is changed before running the apache process

bradymiller:
Just the awesomeness of docker I suppose :)

Only gonna work this into 5.0.2 and flex since I am pretty sure will need to work in a chown apache for the web directory when upgrade docker from 5.0.1 to 5.0.2 in future (gives me time to work out a solution for that which should be easy).

btw, this basically turns the docker into a very nice "data container" for openemr (that even installs automatically!). So can now plug it into any apache/nginx/(even other web servers)/php docker configurations that we want (for example, I am thinking of setting up a newest version apache docker in the dev docker testing environment) or whatever else my limited imagination can not think of :)

jesdynf:
`shadow`?

Also, do you think we should glue all these sequential RUN statements into one layer, if we've got our fingers in here anyways?

bradymiller:
shadow is just to be cool :) (seriously, it is needed to run the `usermod` command).
(btw, I like them separate for now with comments in case run into issues down road)

bradymiller:
To see this in action, click on the following link:
http://play-with-docker.com/?stack=https://gist.githubusercontent.com/bradymiller/fb8f0fbf9fff74b7666a217f0e9c28a1/raw/58e092588b1cb1f51d92e33821b89c8d7a1cbbcd/custom-502-nginx-phpfpm-docker-compose-example.yml

Wait a couple minutes:
Port 80 is nginx with php 7.1
Port 81 is nginx with php 7.2
Port 82 is apache/php (the standard docker use)
(note they are all using the same openemr instance: https://gist.github.com/bradymiller/fb8f0fbf9fff74b7666a217f0e9c28a1 )

bradymiller:
This is testing very well, and going to bring into codebase (the next step I plan on addressing is ensuring that upgrade from 5.0.1 to 5.0.2 does not cause problems).

PR #145, opened 2018-07-04
added redis support to 5.0.2, flex, and flex-edge dockers

bradymiller:
@jesdynf and @TheToolbox ,
I added redis support to the dockers, and is testing well. Note we can't do this to the 5.0.1 docker until after the next patch (5.0.1 breaks when redis php extension is installed and not used locally; actually is a doctrine bug; it will be addressed in patch 4 for 5.0.1)

### DIFF
--- a/docker/openemr/5.0.2/Dockerfile
+++ b/docker/openemr/5.0.2/Dockerfile
@@ -6,7 +6,7 @@ RUN apk add --no-cache \
     apache2 apache2-ssl git php7 php7-tokenizer php7-ctype php7-session php7-apache2 \
     php7-json php7-pdo php7-pdo_mysql php7-curl php7-ldap php7-openssl php7-iconv \
     php7-xml php7-xsl php7-gd php7-zip php7-soap php7-mbstring php7-zlib \
-    php7-mysqli php7-sockets php7-xmlreader perl mysql-client tar curl imagemagick-dev \
+    php7-mysqli php7-sockets php7-xmlreader php7-redis perl mysql-client tar curl imagemagick-dev \
     python openssl git libffi-dev py-pip python-dev build-base openssl-dev dcron \
     shadow
 # Needed to ensure permissions work across shared volumes with openemr, nginx, and php-fpm dockers
--- a/docker/openemr/5.0.2/autoconfig.sh
+++ b/docker/openemr/5.0.2/autoconfig.sh
@@ -131,10 +131,13 @@ if [ "$CONFIG" == "1" ] &&
     fi
 fi
 
-if [ "$REDIS_SERVER" != "" ]; then
+if [ "$REDIS_SERVER" != "" ] &&
+   [ ! -f /etc/php-redis-configured ]; then
     # Variable for $REDIS_SERVER is usually going to be something like 'redis'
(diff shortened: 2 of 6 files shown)


jesdynf:
I wondered how you were cracking that -- it's why I asked @TheToolbox for the ability to launch the image *without* doing any configuration at all. A production deployment that expected to be multi-node would need to run *one* singular configuration instance, complete, and only then let the swarm attack, importing the sqlconf.php from the volume the cluster shares.

That's super-annoying to /do/ in Docker, which is why I haven't either (a) generated a multi-instance Docker-based AWS package or (b) moved Full Stack off the manual installation sequence, which can survive multi-node instantiation as long as only one person is trying to run the initial config.

Well, and also it wasn't smart enough to autoconfigure Redis, so this /is/ progress.

bradymiller:
yep, I'll be able to solve the redis issue easily. Probably the safest thing for a multicontainer setup is to copy what was done for the nginx/php-fpm docker swarm where the openemr docker was simply a data container that autoinstalled. In this case, although seem super weird, I think we could still have an container that is autoinstalled, but have another openemr docker container that shares the html volume of the data container which can be multiplied (with a setting like manual that allows no stuff to happen when they are brought up). Anyways, just thinking out loud. Will see how it pans out in real life.

jesdynf:
I see what you've done there, and I agree with it in principle, but I'd still like that Redis config stanza to set a `./redis-configured ` semaphore somewhere and look for it before running that stanza of code a second time.

jesdynf:
Or am I misreading it and it's impossible for that section to be called twice on the same container (like during a boot)?

bradymiller:
Nope, you are not misreading that and it will run anytime the docker is stop/started. We could plop a dummy file in a non-shared volume spot and use this as a flag for each replication instance to know if it has been booted up yet. Sounds good?

jesdynf:
`/etc/php-redis-configured` would work fine, I've put other semaphores in `/etc` for similar reasons.

bradymiller:
@jesdynf ,
Thanks for the input. Made your recommended changes and the dockers are testing well.
-brady

bradymiller:
After this is in, gonna work on how to make the 5.0.2 docker work in a swarm (without needing the flex docker crutch).

jesdynf:
For what it's worth, I've resigned myself to needing to do that in a separate step, because 'wait for this container to stop before proceeding' is antithetical to how docker-compose thinks. My current plan for the next version of full stack -- or anything load balanced -- needs to spawn an instance and run the configuration under controlled conditions before letting the no-configuration mount-and-go cluster start running. If you can manage to do better, I'll be delighted.

bradymiller:
Thinking ahead on the 5.0.2 swarm safe docker. Am thinking of looking for sites/default/kickoff directory (this needs to be in a shared volume) and if not there:
touch the file there
drop a file /etc/docker-leader

and if is there, do nothing.

Then the instance with the /etc/docker-leader will do everything and the others will do nothing. Again, will see what happens to this in real life.

bradymiller:
Testing is solid, so gonna bring this in and start work on phase 2 :)

PR #150, opened 2018-07-17
support package building in the flex dockers

bradymiller:
Note had to add 2 more php7 packages that were needed for composer run to work(php7-simplexml php7-xmlwriter; will need to bring this into 5.0.2 docker at some point). Also will be no need for bower or npm/nodejs packages (this will be run via a composer package)(will need to remove these and add composer support to the 5.0.2 docker at some point).


bradymiller:
I still need to test this :)

bradymiller:
@jesdynf and @TheToolbox ,
fyi, this is to support packing building of third party packages (so, rather than have them in the official codebase, they will be built during package builds); this is still in the early phases with the goal of testing it in these flex dockers before making it official.

bradymiller:
testing well and will bring in :)

PR #152, opened 2018-07-28
remove sql_patch.php after install since is now included in the patches

bradymiller:


bradymiller:
@jesdynf and @TheToolbox ,
Just removing sql_patch.php from installs since this is now included in the patch file. It's considered a low level security issue since it outputs the version of OpenEMR. No reason to update packages or anything like that.

PR #155, opened 2018-09-03
Prevent local docker volume mount from messing with host permissions

d3sandoval:
adds a new no_chmod env variable that prevents chmod from running during build. This allows for users to mount local volumes safely without worrying about losing access to their host directory.

### DIFF
--- a/docker/openemr/flex-3.7/autoconfig.sh
+++ b/docker/openemr/flex-3.7/autoconfig.sh
@@ -11,6 +11,7 @@
 #  - Optional settings for auto installation are:
 #    - Setting db parameters MYSQL_USER, MYSQL_PASS, MYSQL_DATABASE
 #    - Setting openemr parameters OE_USER, OE_PASS
+#    - EASY_DEV_MODE with value of 'yes' prevents issues with permissions when mounting volumes
 set -e
 
 swarm_wait() {
@@ -45,7 +46,9 @@ auto_setup() {
         CONFIGURATION="${CONFIGURATION} iuserpass=${OE_PASS}"
     fi
 
-    chmod -R 600 /var/www/localhost/htdocs/openemr
+    if [ "$EASY_DEV_MODE" != "yes" ]; then
+        chmod -R 600 /var/www/localhost/htdocs/openemr
+    fi
     php /var/www/localhost/htdocs/auto_configure.php -f ${CONFIGURATION} || return 1
 
     echo "OpenEMR configured."
@@ -144,13 +147,13 @@ if [ -f /etc/docker-leader ] ||
            [ "$SWARM_MODE" == "yes" ]; then
             touch openemr/sites/default/docker-initiated
         fi
-        rsync --recursive --links --exclude .git openemr /var/www/localhost/htdocs/
+        rsync --ignore-existing --recursive --links --exclude .git openemr /var/www/localhost/htdocs/
         rm -fr openemr
         cd /var/www/localhost/htdocs/
     fi
 
     if [ -f /var/www/localhost/htdocs/auto_configure.php ] &&
-       [ ! -d /var/www/localhost/htdocs/openemr/vendor ] &&
+       [[ ! -d /var/www/localhost/htdocs/openemr/vendor || \( -d /var/www/localhost/htdocs/openemr/vendor &&  -z "$(ls -A /var/www/localhost/htdocs/openemr/vendor)" \) ]]  &&
        [ "$FORCE_NO_BUILD_MODE" != "yes" ]; then
         cd /var/www/localhost/htdocs/openemr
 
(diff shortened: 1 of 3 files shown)


docker/openemr/flex-edge/autoconfig.sh
```diff
@@ -144,13 +146,14 @@ if [ -f /etc/docker-leader ] ||
...
-        rsync --recursive --links --exclude .git openemr /var/www/localhost/htdocs/
+        rsync --ignore-existing --recursive --links --exclude .git openemr /var/www/localhost/htdocs/
         rm -fr openemr
         cd /var/www/localhost/htdocs/
     fi
 
     if [ -f /var/www/localhost/htdocs/auto_configure.php ] &&
-       [ ! -d /var/www/localhost/htdocs/openemr/vendor ] &&
+       [ -d /var/www/localhost/htdocs/openemr/vendor ] &&
+       [ -z "$(ls -A /var/www/localhost/htdocs/openemr/vendor)" ]  &&
```
bradymiller: What are the above 2 lines meant to do?

d3sandoval: "if the vendor directory exist and it is empty"
When i was mounting the vendor directory, it mounted an empty dir. so it was skipping these build steps.

It might be better if we say "if the vendor directory doesn't exist or if the vendor directory is empty" but I couldn't figure out how to write that in bash...

bradymiller: But won't this break is a user is using this and has an empty vendor directory?
If so, then need to ensure it works for both cases:
1. A nonexistent vendor directory
2. An empty vendor directory

bradymiller: regarding above, meant to say:
But won't this break is a user is using this and has no vendor directory?

d3sandoval: Yeah... If you can craft an if statement that'll check both cases, I'd be grateful. I wasn't able to figure it out when I was working on it.

bradymiller: try:
```
    if [ -f /var/www/localhost/htdocs/auto_configure.php ] &&
       [[ ! -d /var/www/localhost/htdocs/openemr/vendor || ( -d /var/www/localhost/htdocs/openemr/vendor  && -z "$(ls -A /var/www/localhost/htdocs/openemr/vendor)" ) ]]  &&
       [ "$FORCE_NO_BUILD_MODE" != "yes" ]; then
```
I have no idea if it will work, but adapted it from here: https://stackoverflow.com/questions/14964805/groups-of-compound-conditions-in-bash-test

jesdynf: https://stackoverflow.com/questions/3427872/whats-the-difference-between-and-in-bash for a little more insight into what's going on there.

bradymiller:
It looks like the NO_CHMOD does more than not allow no chmod. Would it make sense to label it a setting like `EASY_DEV_MODE` since it is specific to setting up the Easy Development Environment Mode?

docker/openemr/flex-edge/autoconfig.sh
bradymiller: The above should work (fingers crossed) :)

jesdynf:
FYI, the failing check is probably coming from the mysql docker which I've patched.

docker/openemr/flex-edge/autoconfig.sh
d3sandoval: Close! Had to escape the parens:
```bash
    if [ -f /var/www/localhost/htdocs/auto_configure.php ] &&
       [[ ! -d /var/www/localhost/htdocs/openemr/vendor || \( -d /var/www/localhost/htdocs/openemr/vendor &&  -z "$(ls -A /var/www/localhost/htdocs/openemr/vendor)" \) ]]  &&
       [ "$FORCE_NO_BUILD_MODE" != "yes" ]; then
```

See: https://stackoverflow.com/a/3826462/3164104

docker/openemr/flex-3.7/autoconfig.sh
```diff
@@ -183,12 +186,14 @@ if [ -f /etc/docker-leader ] ||
...
 
-    if [ -f /var/www/localhost/htdocs/auto_configure.php ]; then
+    if [ -f /var/www/localhost/htdocs/auto_configure.php ] &&
+       [ "$EASY_DEV_MODE" != "yes" ]; then
         chmod 666 /var/www/localhost/htdocs/openemr/sites/default/sqlconf.php
         chmod 666 /var/www/localhost/htdocs/openemr/interface/modules/zend_modules/config/application.config.php
-        chown -R apache /var/www/localhost/htdocs/openemr/
     fi
 
+    chown -R apache /var/www/localhost/htdocs/openemr/
```
bradymiller: hi @d3sandoval ,
To be safe, this line should only run if /var/www/localhost/htdocs/auto_configure.php exists (to avoid running it again and again on docker compose stop/start). I am happy to make this change, just want to make sure you don't see it breaking anything for your stuff.

bradymiller: (just let me know; if it's ok, then I'll merge in this PR and then make my change on a commit following)

d3sandoval:
@bradymiller that change is fine as long as it still runs when EASY_DEV_MODE is set to "yes". Without it, Apache can't serve webpages

PR #156, opened 2018-10-25
Minor Ehrlich Updates

danehrlich1:
1. Some changes to php.ini

- hardened some variables related to PHP sessions

2. Added httpd.conf

- can replace this with awk/sed commands later if liked 
- adds a series of  secure HTTPS headers to any web server responses
- also turns  off Server Signature, Server  Tokens, and Server ETags for further security

### DIFF
--- a/docker/openemr/5.0.2/Dockerfile
+++ b/docker/openemr/5.0.2/Dockerfile
@@ -51,6 +51,8 @@ COPY utilities/unlock_admin.php utilities/unlock_admin.sh /root/
 RUN chmod 500 run_openemr.sh autoconfig.sh /root/unlock_admin.sh \
     && chmod 000 auto_configure.php /root/unlock_admin.php
 #fix issue with apache2 dying prematurely
+RUN rm -f /etc/apache2/httpd.conf
+COPY httpd.conf /etc/apache2/
 RUN mkdir -p /run/apache2
 #go
 CMD [ "./run_openemr.sh" ]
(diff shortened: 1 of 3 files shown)


docker/openemr/5.0.2/Dockerfile
```diff
@@ -51,6 +51,8 @@ COPY utilities/unlock_admin.php utilities/unlock_admin.sh /root/
 RUN chmod 500 run_openemr.sh autoconfig.sh /root/unlock_admin.sh \
     && chmod 000 auto_configure.php /root/unlock_admin.php
 #fix issue with apache2 dying prematurely
+RUN rm -f /etc/apache2/httpd.conf
```
jesdynf: Your COPY directive will overwrite the target -- you may remove this line.

docker/openemr/5.0.2/php.ini
```diff
@@ -1345,14 +1345,14 @@ session.save_handler = files
 ; applications from session fixation via session adoption vulnerability. It is
 ; disabled by default for maximum compatibility, but enabling it is encouraged.
 ; https://wiki.php.net/rfc/strict_sessions
-session.use_strict_mode = 0
+session.use_strict_mode = 1
```
jesdynf: @bradymiller Is this permissible? Has OpenEMR been properly tested for compatibility with this flag?

bradymiller: hi @jesdynf , I need to do some research on this. -brady

bradymiller:
Agree that we need to see the http.conf changes from the stock. to do this @danehrlich1  , do the following. Copy over the stock http.conf and then make that a commit. And then copy back over your modified http.conf and then make another commit. Then we will be able to look at your last commit and see the changes.

danehrlich1:
Closing this. Will make some kind of a pull request for later.

PR #157, opened 2018-11-16
Orchestrations

danehrlich1:
`docker-compose.yml` that includes ModSecurity Firewall / Reverse Proxy (still testing a lot)

Basic Kubernetes version of OpenEMR 

Will upload instructions in additional commit


docker/openemr/5.0.2/Dockerfile
```diff
@@ -51,6 +51,8 @@ COPY utilities/unlock_admin.php utilities/unlock_admin.sh /root/
 RUN chmod 500 run_openemr.sh autoconfig.sh /root/unlock_admin.sh \
     && chmod 000 auto_configure.php /root/unlock_admin.php
 #fix issue with apache2 dying prematurely
```
bradymiller: above comment should be above the RUN mkdir line below

danehrlich1:
@jesdynf Ignore some of that stuff. I'm working with @bradymiller to clean up all my branches and pull requests on here and the regular openemr repo currently.

The 5.0.2u you're seeing...the "U" stood for Ubuntu at one time.

Anyway I need a few more weeks of working on Docker to understand and get all  the images built, but at the end of it all I will have a container running OpenEMR, based off of your image, that has:
 - ModSecurity WAF Layer 7 firewall - compiled and running directly in Apache
 - GeoIP Country blocking - from MaxMind so users can lock down their instance to a specific country(s)
 - IP blacklist from Emerging Threats - updated daily with the worst 10,000 IP's on the internet

PR #186, opened 2019-03-13
Fixed bugs in Openemr Apache conf file

danehrlich1:
GH-185

Need to test.

Removed `allowsymlinks` for now...maybe can be replaced by `allowsymlinksifownersmatch` but not sure. See this pretty shocking post: https://blog.sucuri.net/2013/05/from-a-site-compromise-to-full-root-access-symlinks-to-root-part-i.html

Only `html2pdf` uses `symlink` php function...once, will try to find a way to refactor this out and then restrict in `php.ini`.

Added:
HTTPProtocolOptions Strict
SSLHonorCipherOrder
Refactored SSLCipherSuite

### DIFF
--- a/docker/openemr/5.0.2/openemr.conf
+++ b/docker/openemr/5.0.2/openemr.conf
@@ -1,29 +1,24 @@
 LoadModule rewrite_module modules/mod_rewrite.so
 LoadModule allowmethods_module modules/mod_allowmethods.so
 
-
-# The ServerName directive sets the request scheme, hostname and port that
-# the server uses to identify itself. This is used when creating
-# redirection URLs. In the context of virtual hosts, the ServerName
-# specifies what hostname must appear in the request's Host: header to
-# match this virtual host. For the default virtual host (this file) this
-# value is not decisive as it is used as a last resort host regardless.
-# However, you must set it for any further virtual host explicitly.
-#ServerName www.example.com
-
-
 ## Security Options
+# Strong HTTP Protocol
+HTTPProtocolOptions Strict
 Protocols http/1.1
-# Removes revealing HTTP Headers
+        
+# Don't Reveal Server
 ServerSignature off
 ServerTokens Prod
-TraceEnable Off
+Header unset Server
+
+# No ETag
 FileETag None
 Header unset ETag
+
 # Set HSTS and X-XSS protection
 Header set Strict-Transport-Security "max-age=31536000; includeSubDomains; preload"
 Header set X-XSS-Protection "1; mode=block"
-ServerAdmin webmaster@localhost
+
 # Narrow document root
 DocumentRoot /var/www/localhost/htdocs/openemr
 ErrorLog ${APACHE_LOG_DIR}/error.log
@@ -31,10 +26,8 @@ CustomLog ${APACHE_LOG_DIR}/access.log combined
 
 
     <Directory /var/www/localhost/htdocs/openemr>
-        # Only allow these HTTP Methods
-        AllowMethods GET POST HEAD OPTIONS
-        # No indexes anywhere
-        Options -Indexes
+        AllowMethods GET POST #Only allow these HTTP Methods
+        Options None 
         AllowOverride FileInfo
         Require all granted
     </Directory>
@@ -58,8 +51,8 @@ CustomLog ${APACHE_LOG_DIR}/access.log combined
 
 #######################################
 ### Uncomment the following 3 lines ###
-### to enable HTTPS redirection #######
-### and require HTTPS only ############
+### with #'s below to enable HTTPS  ###
+### redirection & require HTTPS only ##
 #######################################
 <VirtualHost *:80>
         #RewriteEngine On
@@ -69,13 +62,9 @@ CustomLog ${APACHE_LOG_DIR}/access.log combined
 
 
 <VirtualHost _default_:443>
-        #   SSL Engine Switch:
-        #   Enable/Disable SSL for this virtual host.
         SSLEngine on
-
-        SSLCipherSuite "EECDH+ECDSA+AESGCM EECDH+aRSA+AESGCM EECDH+ECDSA+SHA384 \
-                       EECDH+ECDSA+SHA256 EECDH+aRSA+SHA384 EECDH+aRSA+SHA256 EECDH+aRSA+RC4 \
-                       EECDH EDH+aRSA RC4 !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS"
+        SSLHonorCipherOrder on
+        SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH
         SSLProtocol -ALL +TLSv1.2
         SSLCertificateFile    /etc/ssl/certs/webserver.cert.pem
         SSLCertificateKeyFile /etc/ssl/private/webserver.key.pem


jesdynf:
Daniel, I'm sorry this sat for so long; can you patch this so it merges in? And @bradymiller , can you verify these will be non-problems? I'm worried that we might need an Apache option and that we're just -Indexes for a reason.

docker/openemr/5.0.2/openemr.conf
```diff
@@ -1,40 +1,33 @@
...
 ErrorLog ${APACHE_LOG_DIR}/error.log
 CustomLog ${APACHE_LOG_DIR}/access.log combined
 
 
     <Directory /var/www/localhost/htdocs/openemr>
-        # Only allow these HTTP Methods
-        AllowMethods GET POST HEAD OPTIONS
-        # No indexes anywhere
-        Options -Indexes
+        AllowMethods GET POST #Only allow these HTTP Methods
```
bradymiller: Should not do this line (note I have had to add PUT and DELETE to this so far for the api support). Guessing that HEAD and OPTIONS can be removed though. thoughts @danehrlich1 ?

docker/openemr/5.0.2/openemr.conf
```diff
@@ -1,40 +1,33 @@
...
-# specifies what hostname must appear in the request's Host: header to
-# match this virtual host. For the default virtual host (this file) this
-# value is not decisive as it is used as a last resort host regardless.
-# However, you must set it for any further virtual host explicitly.
-#ServerName www.example.com
-
-
 ## Security Options
+# Strong HTTP Protocol
+HTTPProtocolOptions Strict
```
bradymiller: Not sure what this does, but can always try it out. @danehrlich1 , this won't force https, correct?

docker/openemr/5.0.2/openemr.conf
```diff
@@ -1,40 +1,33 @@
...
+Header unset Server
+
+# No ETag
 FileETag None
 Header unset ETag
+
 # Set HSTS and X-XSS protection
 Header set Strict-Transport-Security "max-age=31536000; includeSubDomains; preload"
 Header set X-XSS-Protection "1; mode=block"
-ServerAdmin webmaster@localhost
```
bradymiller: this line makes sense

danehrlich1:
Yes HEAD can be removed. It’s redundant since GET is interpreted by Apache to be either GET or HEAD confusingly.

You tell me in regards to OPTIONS :)

docker/openemr/5.0.2/openemr.conf
```diff
@@ -1,40 +1,33 @@
...
 ## Security Options
+# Strong HTTP Protocol
+HTTPProtocolOptions Strict
 Protocols http/1.1
-# Removes revealing HTTP Headers
+        
+# Don't Reveal Server
 ServerSignature off
 ServerTokens Prod
-TraceEnable Off
```
bradymiller: @danehrlich1 , just checking why remove this line?

danehrlich1:
No it forces HTTP 1.1, which came out in 1999.

History of HTTP is divided into 3 phases (before HTTP2 but that’s another story).

0.9 - 1989
1.0 - 1996
1.1  - 1999

The just forces you to use HTTP 1.1. Reasonable request and better security.

danehrlich1:
I dunno. Maybe accident? Is Trace Enable set somewhere else so it was duplicative?

docker/openemr/5.0.2/openemr.conf
```diff
@@ -1,40 +1,33 @@
...
+Header unset Server
```
bradymiller: This line looks good, but appears it may not do anything:
https://stackoverflow.com/questions/35360516/cant-remove-server-apache-header

docker/openemr/5.0.2/openemr.conf
```diff
@@ -1,40 +1,33 @@
...
 CustomLog ${APACHE_LOG_DIR}/access.log combined
 
 
     <Directory /var/www/localhost/htdocs/openemr>
-        # Only allow these HTTP Methods
-        AllowMethods GET POST HEAD OPTIONS
-        # No indexes anywhere
-        Options -Indexes
+        AllowMethods GET POST #Only allow these HTTP Methods
+        Options None
```
bradymiller: Pretty sure Indexes should not be required anywhere, so this makes sense

docker/openemr/5.0.2/openemr.conf
```diff
@@ -69,13 +62,9 @@ CustomLog ${APACHE_LOG_DIR}/access.log combined
...
 <VirtualHost _default_:443>
-        #   SSL Engine Switch:
-        #   Enable/Disable SSL for this virtual host.
         SSLEngine on
-
-        SSLCipherSuite "EECDH+ECDSA+AESGCM EECDH+aRSA+AESGCM EECDH+ECDSA+SHA384 \
-                       EECDH+ECDSA+SHA256 EECDH+aRSA+SHA384 EECDH+aRSA+SHA256 EECDH+aRSA+RC4 \
-                       EECDH EDH+aRSA RC4 !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS"
+        SSLHonorCipherOrder on
+        SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH
```
bradymiller: Got know idea what above SSL changes to but guessing @danehrlich1 did these for good reasons.

danehrlich1:
NOOOOOOOOOOOOOOOOOOO

Damn Apache can never do anything right. Lol so many directives that don’t even work.

bradymiller:
Agree on the Option issue, but don't know. Looks like this was still in testing phase by @danehrlich1 and there are merge conflicts. Seem best if @danehrlich1 can possibly resubmit changes based on most recent code.

bradymiller:
oh goodness, here's a light read on apache options:
https://httpd.apache.org/docs/2.4/mod/core.html#options
:)

docker/openemr/5.0.2/openemr.conf
bradymiller: edit; i misunderstood this line. agree that unsure what will happen if remove all Options.

docker/openemr/5.0.2/openemr.conf
bradymiller: Lets just ensure we aren't removing any modern browser compatibilities with this more restricted  SSLCipherSuite

bradymiller:
hi @jesdynf and @danehrlich1 , Left some comments. -brady

jesdynf:
For what it's worth, the OPTIONS verb matters when CORS is enabled, but I don't know how OpenEMR's session state interacts with CORS; it may or may not be necessary.

danehrlich1:
We can leave it I see no problem.

The real security issues with Apache, and these I still have never found an answer for, are probably:
- followsymlinks (unsolved problem)
- making sure the file permissions are correct (very hard to ever nail down 100%)
- probably something with mimetypes and character encoding later on

docker/openemr/5.0.2/openemr.conf
bradymiller: Looked this up and see why you removed this line (apache docs state this is not a security vulnerability)

docker/openemr/5.0.2/openemr.conf
bradymiller: btw, regarding HEAD, appears good idea to keep for caching of static files (like assets, for example)

docker/openemr/5.0.2/openemr.conf
bradymiller: Does something like this make more sense:
https://httpd.apache.org/docs/2.4/ssl/ssl_howto.html#onlystrong

bradymiller: Some good reading on this topic:
https://stackoverflow.com/questions/28737374/sslciphersuite-aliases
https://blog.qualys.com/ssllabs/2011/10/17/mitigating-the-beast-attack-on-tls

bradymiller:
made changes and migrated this PR to the following PR:
https://github.com/openemr/openemr-devops/pull/193

PR #191, opened 2019-05-04
support for new writable requirements

bradymiller:


bradymiller:
hi @adunsulag ,
I just realize that if I update the flex dockers with these changes that they will break when they are used on the older codebase (the current flex dockers without these changes work for both old and new codebase, but then not going to test a unwritable application.config.php). Would it make sense to wait a bit on this or even adding a docker parameter to manually support older codebase?

adunsulag:
@bradymiller What are you thinking in terms of waiting?  

Do we just want to update the flex-edge and leave the flex3.9 alone?  So people can code against that version?  Seems like flex-edge should be going against the latest and greatest and people who use that should understand they'll have to merge any changes from master into whatever branch they are coding against.  If they want to be compatible w/ the earlier docker they can use the flex3.9 version.  Am I off in this?

bradymiller:
That seems reasonable. Will make the change for 5.0.2 and flex-edge now and then do flex (flex3.9) after we release 5.0.2 in the future.

bradymiller:
Going to bring this in (5.0.2 and flex-edge) and build the new dockers now. And will then make a PR to use in future (for edge-3.9 docker) which will use in future after release of 5.0.2.

bradymiller:
looks like picked a bad day to try to update the docker :)
https://success.docker.com/article/docker-hub-maintenance

PR #193, opened 2019-05-12
update apache settings in dockers

bradymiller:
this is from the prior PR:
https://github.com/openemr/openemr-devops/pull/186

to see the code without whitespace changes, use following link:
https://github.com/openemr/openemr-devops/pull/193/files?w=1


bradymiller:
updated dockers which are testing well, so bringing this in.

PR #208, opened 2019-09-05
Xdebug

boxlady:
I have added xdebug, so that I can connect to docker and do step debugging. 

I have also changed the line ending in the shell scripts, however git ignores those changes. In order to make those changes to everyone I added .gitattributes for check out and check in.

Here is a link for more info
https://help.github.com/en/articles/configuring-git-to-handle-line-endings


jesdynf:
I approve in principle, but can @bradymiller confirm that the -xdebug stuff is unlikely to get accidentally copied into production dockerfiles? I don't know how those get made for new versions.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Shell scripts
+*.sh text eol=lf
```
bradymiller: To clarify, what OS were your builds breaking on?

docker/openemr/flex-3.10/php.ini
```diff
@@ -1947,3 +1947,12 @@ ldap.max_links = -1
...
 ; End:
+[Xdebug]
+zend_extension=/usr/lib/php7/modules/xdebug.so
+xdebug.coverage_enable=0 
+xdebug.remote_enable=1      
+xdebug.remote_connect_back=0
+xdebug.remote_log=/tmp/xdebug.log
+xdebug.remote_autostart=1  
+xdebug.remote_port=9001
+xdebug.idekey = PHPSTORM
```
bradymiller: If we were to make this configurable as a docker environment variable(s), which settings above should be configurable for the developer? Once I know this, then can provide guidance on making this a docker environment settings. To get an idea of what I mean, check out how the `REDIS_SERVER` environment setting works.

.gitattributes
boxlady: My OS is windows, and the container was Linux. From what I understand, I pulled it to my OS and the line endings where changed. When I tried to run it failed because it was the line endings used in windows instead of the line endings needed for Linux.

bradymiller: Thanks for the info. Makes sense since need to keep the linux line endings since they are being copied into a linux docker.

docker/openemr/flex-3.10/php.ini
boxlady: I think Port, and Idkey, and remote_host are three of the main ones.  This is what I added to the compose file.
         ports: 9000:9001
 environment:    
            XDEBUG_CONFIG: remote_host=host.docker.internal
            PHP_IDE_CONFIG: "serverName=localhost"
       
I think this explains it pretty well.
https://serversforhackers.com/s/xdebug-in-docker

.gitattributes
bradymiller: Would expand on your comment, so we don't forget what it is for:
```
# Shell scripts need to maintain linux line endings even in windows environments (since they are being copied into linux docker environments).
```

boxlady:
@bradymiller  I have added a new folder/image so that it doesn't interfere with others. I have also added a guide in the readme on how to set it up. with some images. Let me know if you have any questions.

bradymiller:
just added support for xdebug and profiling in the flex dockers here:
https://github.com/openemr/openemr-devops/commit/03b8208bfc91bb5d1a92bea43931e0dc8118ba13

bradymiller:
btw, instructions for use are in step 10 of https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md

PR #210, opened 2019-10-15
Rsync fails without openemr naming convention

generalconsensus:
In order for openemr to work with nonstandard names it requires that the cloned repo have the name requested in the rsync statement or the image fails

https://github.com/openemr/openemr-devops/blob/master/docker/openemr/flex-3.10/autoconfig.sh#L150


jesdynf:
This is the flex development-only docker -- under what circumstances should you be running an rsync against it?

PR #225, opened 2020-01-25
easy development docker environment fix

bradymiller:
It seem like the easy development docker should use the local repo code rather than grabbing code from a online repository. Changed it to do that. Still need to build and test this before bringing in.


tywrenn:
This is now of high priority to fix due to me not able to view changes.

bradymiller:
Working on this now. Getting weird missing file issues on testing. I should get it sorted out soon hopefully.

bradymiller:
I think i know the issue here and why the original author of this docker-compose had to go this route. When you make a shared volume within a shared volume the inner one is always blank, so then by grabbing the code from github, it would refill in these blank shared volumes. There's gotta be another way.

tywrenn:
Found a fix for my issue yet it might be wise to fix this anyway.

bradymiller:
testing well. bringing this in and building/updating dockers and the docker-compose.yml

PR #230, opened 2020-03-24
WIP: Add php developer tools to openemr flex3.1

### LINKED ISSUE #228: Add PHP Developer tools to docker environment
### Description
Hello @bradymiller, This is in reference to the issue discussed [here](https://github.com/openemr/openemr/issues/3193). I was thinking a good starting point would be creating an issue to bring in the developer dependencies here. 

Add support to the development docker to allow the developer to do the following in the docker:

php syntax checking (DONE)
psr2 checking/fixing (DONE)
unit tests (DONE)
e2e tests (DONE)
api tests (DONE)

Once you confirm this is the way to go, I would add the scripts as you described in the issue above

prondubuisi:
Fixes #228

### DIFF
--- a/docker/openemr/flex-3.10/autoconfig.sh
+++ b/docker/openemr/flex-3.10/autoconfig.sh
@@ -167,7 +167,12 @@ if [ -f /etc/docker-leader ] ||
 
         # install php dependencies
         composer install
-
+        if [ "$DEVELOPER_TOOLS" == "yes" ]; then
+            composer global require "squizlabs/php_codesniffer=3.*" 
+            composer require "phpunit/phpunit=8.*"
+            composer require "symfony/panther=^0.6" 
+        fi
+        
         if [ -f /var/www/localhost/htdocs/openemr/package.json ]; then
             # install frontend dependencies (need unsafe-perm to run as root)
             npm install --unsafe-perm
--- a/docker/openemr/flex-3.11/autoconfig.sh
+++ b/docker/openemr/flex-3.11/autoconfig.sh
@@ -167,6 +167,11 @@ if [ -f /etc/docker-leader ] ||
 
         # install php dependencies
         composer install
+        if [ "$DEVELOPER_TOOLS" == "yes" ]; then
+            composer global require "squizlabs/php_codesniffer=3.*" 
+            composer require "phpunit/phpunit=8.*"
+            composer require "symfony/panther=^0.6" 
+        fi
 
         if [ -f /var/www/localhost/htdocs/openemr/package.json ]; then
             # install frontend dependencies (need unsafe-perm to run as root)
--- a/docker/openemr/flex-edge/autoconfig.sh
+++ b/docker/openemr/flex-edge/autoconfig.sh
@@ -167,6 +167,11 @@ if [ -f /etc/docker-leader ] ||
 
         # install php dependencies
         composer install
+        if [ "$DEVELOPER_TOOLS" == "yes" ]; then
+            composer global require "squizlabs/php_codesniffer=3.*" 
+            composer require "phpunit/phpunit=8.*"
+            composer require "symfony/panther=^0.6" 
+        fi
 
         if [ -f /var/www/localhost/htdocs/openemr/package.json ]; then
             # install frontend dependencies (need unsafe-perm to run as root)


prondubuisi:
Was Wondering if I need to push my local addition here https://github.com/openemr/openemr/blob/master/docker-compose.yml#L48 in my local branch to OpenEMR? i.e  DEVELOPER_TOOLS: "yes"

docker/openemr/flex-3.10/autoconfig.sh
```diff
@@ -167,7 +167,15 @@ if [ -f /etc/docker-leader ] ||
 
         # install php dependencies
         composer install
+        if [ "$DEVELOPER_TOOLS" == "yes" ]; then
+            composer global require "squizlabs/php_codesniffer=3.*" 
+            composer require "phpunit/phpunit=8.*"
+            composer require "symfony/panther=^0.6"
```
bradymiller: 2DV

docker/openemr/flex-3.10/autoconfig.sh
```diff
@@ -167,7 +167,15 @@ if [ -f /etc/docker-leader ] ||
 
         # install php dependencies
         composer install
+        if [ "$DEVELOPER_TOOLS" == "yes" ]; then
+            composer global require "squizlabs/php_codesniffer=3.*" 
+            composer require "phpunit/phpunit=8.*"
+            composer require "symfony/panther=^0.6" 
 
+             # optimize
+             composer dump-autoload -o
```
bradymiller: line 200 already does this, so can drop above 2 lines

bradymiller:
definitely a yes on pushing that change to https://github.com/openemr/openemr/blob/master/docker-compose.yml#L48

Also, make the same change in
https://github.com/openemr/openemr-devops/blob/master/docker/openemr/flex-3.11/autoconfig.sh
https://github.com/openemr/openemr-devops/blob/master/docker/openemr/flex-edge/autoconfig.sh

Then will bring in this PR and I'll rebuild the dockers and push them out.

docker/openemr/flex-3.10/autoconfig.sh
prondubuisi: Lol super excited.

bradymiller:
btw, ignore travis here.

docker/openemr/flex-3.10/autoconfig.sh
prondubuisi: Done.

bradymiller:
hi @prondubuisi ,
Looks good and bringing this in. I'll build dockers (do a quick sanity test) and then push them to dockerhub. thanks for this improvement, which I am sure will decrease the frequency of t-PTSD (travis post traumatic stress disorder)!

prondubuisi:
> hi @prondubuisi ,
> Looks good and bringing this in. I'll build dockers (do a quick sanity test) and then push them to dockerhub. thanks for this improvement, which I am sure will decrease the frequency of t-PTSD (travis post traumatic stress disorder)!

Lol. Welcome @bradymiller. Would love to know when you make the push. Also what does a sanity test entail?

bradymiller:
basically just running the docker.
Turns out we have a very annoying issue to deal with. The composer require (not the global one) for a token check to use the github api to check on the vcs repository for wkhtmltopdf-openemr (note openemr does not even install this by default but the github api still forces a check; if the developer has not used up the anonymous limit, then ok, but it's not uncommon to hit that limit and then it basically breaks). Travis setup had this same issue, which we solved by creating a github token for that purpose. But we can't do that here. Two options are:
1. Migrate wkhtmltopdf-openemr to packagist (@sjpadgett , what would doing that entail)
2. Support 'composer install --no-dev' at places in infrastructure where production use happens and then can add the "phpunit/phpunit=8.*" and "symfony/panther=^0.6" to the dev part of composer.json.
Either option will fix this issue. And we should actually do both. I'll research what option 2 will entail and await @sjpadgett advice on option 1.

bradymiller:
btw, we can still support the phpcs since the require global works fine, and that is the critical improvement. I will comment out the other 2 requires with above post as a comment which we can work out in future.

prondubuisi:
> btw, we can still support the phpcs since the require global works fine, and that is the critical improvement. I will comment out the other 2 requires with above post as a comment which we can work out in future.

Cool. Will be here to see how we fix this

PR #236, opened 2020-04-04
Add php error log alias to devtools

### LINKED ISSUE #234: Add Alias for PHP error log on docker
### Describe the Problem
In line with our aliasing other dev tools commands, I think it will be okay to also alias the PHP error log command for consistency. Also a close comparison of the command and proposed alias show that the alias is a bit shorter and easier to commit to memory 

    docker exec -i $(docker ps | grep _openemr | cut -f 1 -d " ") sh -c 'cat /var/log/apache2/error.log'

    ```
    docker exec -i $(docker ps | grep _openemr | cut -f 1 -d " ") sh -c '/root/devtools php-log'

### Describe Solution
  add php error log alias
    docker exec -i $(docker ps | grep _openemr | cut -f 1 -d " ") sh -c '/root/devtools php-log'

prondubuisi:
Fixes #234

### DIFF
--- a/docker/openemr/flex-3.10/utilities/devtools
+++ b/docker/openemr/flex-3.10/utilities/devtools
@@ -1,6 +1,11 @@
 #! /bin/sh
 
 
+if [ "$1" == "php-log" ]; then
+    echo "Generating PHP error log"
+    cat /var/log/apache2/error.log
+fi
+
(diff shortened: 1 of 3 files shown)


bradymiller:
hi @prondubuisi ,
Looks good and bringing this in, so it will be in the next docker build. thanks for the imrovement! -brady

PR #252, opened 2020-05-16
Support setting openemr globals

### LINKED ISSUE #240: Support any OpenEMR globals setting in docker setup
There is now a cool setting where can turn on the API by default as a docker environment(https://github.com/openemr/openemr-devops/blob/master/docker/openemr/flex-3.11/autoconfig.sh#L75). Something that would be really cool would be to take it further and support the setting of any global (ie. setting). This actually wouldn't be very tough to do. Would basically have an environment setting such as GLOBAL_SETTINGS (which is an array with key being the global gl_name (such as `rest_api`) and the value being the `gl_value` setting. Then can set anything you want as a docker environment setting (for example, theme, locale stuff, language) and will take very minimal code to support it.

bradymiller:
fixes #240 

This allows setting any openemr global from the docker-compose.yml . For example, plan to place following environment setting in the easy dev docker environment docker-compose.yml:
```
OPENEMR_SETTING_rest_api: 1
OPENEMR_SETTING_rest_fhir_api: 1
OPENEMR_SETTING_rest_portal_api: 1
OPENEMR_SETTING_rest_portal_fhir_api: 1
OPENEMR_SETTING_portal_onsite_two_enable: 1
```

### DIFF
--- a/docker/openemr/5.0.3/autoconfig.sh
+++ b/docker/openemr/5.0.3/autoconfig.sh
@@ -60,15 +60,18 @@ auto_setup() {
     fi
 
     # Set requested openemr settings
-    echo "`printenv | grep '^OPENEMR_SETTING_'`" |
-    while IFS= read -r line; do
-        SETTING_TEMP=`echo "$line" | cut -d "=" -f 1`
-        # note am omitting the letter O on purpose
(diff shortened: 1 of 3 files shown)


bradymiller:
Testing well and bringing this in. Will update dockers and the docker-compose.yml in main repo.

PR #255, opened 2020-07-11
utf8mb4 support and devtools

bradymiller:
this is companion changes to the following PR:
https://github.com/openemr/openemr/pull/3721


bradymiller:
@asher .
Regarding changing database encoding and collation that we discussed on the call, check out `change-encoding-collation` command stuff in devtools

PR #256, opened 2020-07-13
add openemr-cmd script

reidliu41:
Fixes # 

openemr-cmd may help developer easier to manage docker or docker-compose command
https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md

-
-
-


jesdynf:
Hello, Reid. A few notes:

 * Use --detach for docker-compose instead of redirection, see https://docs.docker.com/compose/reference/up/
 * There's no reason you need the color spec in a grep inside a pipe.
* No provisions are made to leave that while loop if an unexpected condition (like the docker command instantly exiting) prevents the loop from ever exiting.

Basically I don't find the start_ and stop_ commands especially well thought out -- they're not adding anything to the base command, if detach is used, and they're touching secondary parts of your system and they've got an odd sensibility about pathing. The index of insights into devtools seem more immediately useful.

reidliu41:
@jesdynf Asher, thanks for your notes.
I re-think again, docker-compose will the output by default, it's nicer with -d. I get rid of the process check and docker-compose stop it with the "down" command.

And you think no need to use start_ and stop_ part?

reidliu41:
@jesdynf I checked the 'docker-compose up -d' can repeat to execute, nicer. but 'docker-compose down' is fine at the first time. Seems that no need to check process up or down, just run up or down.
```
[root@openemr openemr]# docker-compose up -d
openemr_mysql_1 is up-to-date
openemr_phpmyadmin_1 is up-to-date
openemr_openemr_1 is up-to-date
[root@openemr openemr]# docker-compose up -d
openemr_mysql_1 is up-to-date
openemr_phpmyadmin_1 is up-to-date
openemr_openemr_1 is up-to-date

[root@openemr openemr]# docker-compose down
Stopping openemr_openemr_1    ... done
Stopping openemr_phpmyadmin_1 ... done
Stopping openemr_mysql_1      ... done
Removing openemr_openemr_1    ... done
Removing openemr_phpmyadmin_1 ... done
Removing openemr_mysql_1      ... done
Removing network openemr_default
[root@openemr openemr]# docker-compose down
Removing network openemr_default
WARNING: Network openemr_default not found.
```

jesdynf:
You're closer, but I still think you're overreaching with the start and stop commands. You're trying to establish the dockers as a system-level utility accessible by any user without providing a permanent location they're to be found, so you just check to see if the current location is the right one and then fail if it doesn't work.

I would honestly either remove start and stop entirely or just alias them to...
```
docker-compose up -d
exit $?
```
The bits with /var/lock, the needless echo-reporting, and the process-searching and grepping should go. A script in the current user's /home/*/bin that makes dev commands easily-accessible one-shots is still useful.

bradymiller:
hi @reidliu41 ,

Very cool to see work on this stuff! (shout out to @prondubuisi to making all this devtools stuff possible :) ). I'm plus/minus on the start/stop stuff and agree with @jesdynf , but definitely not a show stopper or anything like that, since users can simply not use it. For `docker-compose down`, note would strongly consider also using the `-v` flag since all the shared volumes generally just break a bunch of things when try to do another `docker-compose up`. When this thing is ready, I'll definitely be using it since it will make calling the devtools command much easier (and will be a good place in future for folks to document what all the devtools do).

Other recommendations would be to:
1. move this script and README.md to new folder in this repo called: /utilities/devtools/
2. make the files and command all lowercase
3. add support for the new devtool that was just added: https://github.com/openemr/openemr-devops/blob/master/docker/openemr/flex-edge/utilities/devtools#L229

bradymiller:
Just noted the capsule command (bringing them back and forth from host and docker). Will be an awesome feature for devs for it to be so simple!
2DV

reidliu41:
@jesdynf @bradymiller thanks for your advice, I already moved the stop_ and start_. Actually, I just wrote from the operation side(try simple), you guys provide me some thought from the dev side, that's great.

docker/openemr/flex-edge/utilities/openemr-cmd/openemr-cmd
```diff
@@ -0,0 +1,302 @@
...
+	echo "  clean-sweep-tests            To run only all the automated tests" 
+	echo "  dev-reset                    To reset OpenEMR only"
+	echo "  dev-reset-install            To reset and reinstall OpenEMR"
+	echo "  dev-reset-install-demodata   To reset and reinstall OpenEMR with demo data"
+	echo "  backup-snapshot              Create a backup snapshot"
+	echo "  restore-snapshot             Restore from a snapshot"
+	echo "  list-snapshots               To list the snapshots"
+	echo "  list-capsules                List the capsules"
+	echo "  cp-cap-docker2dir            Copy the capsule from the docker to your current directory" 
+	echo "  cp-cap-dir2docker            Copy the capsule from current directory into the docker"
```
bradymiller, line 189: in the above 2 commands, don't hardcode example.tgz, and instead make this a parameter in the command, this the command would look something like:
```
openemr-cmd cp-cap-docker2dir example.tgz
```
(could even get fancy and check to ensure the capsule exists before copying it)

docker/openemr/flex-edge/utilities/openemr-cmd/openemr-cmd
```diff
@@ -0,0 +1,302 @@
...
+	echo "  list-capsules                List the capsules"
+	echo "  cp-cap-docker2dir            Copy the capsule from the docker to your current directory" 
+	echo "  cp-cap-dir2docker            Copy the capsule from current directory into the docker" 
+	echo "  ensure-version               Ensure run upgrade to ensure will work with current version OpenEMR"
+	echo "  enable-multisite             Turn on support for multisite"
+	echo "  disable-multisite            Turn off support for multisite"
+	echo "  utf8mb4-general-ci			 Set character set to utf8mb4 and collation to utf8mb4_general_ci"
+	echo "  utf8mb4-unicode-ci           Set character set to utf8mb4 and collation to utf8mb4_unicode_ci"
+	echo "  utf8mb4-vietnamese-ci		 Set character set to utf8mb4 and collation to utf8mb4_vietnamese_ci"
+	echo "  utf8-general-ci              Set character set to utf8 and collation to utf8_general_ci"
```
bradymiller, line 196: Reduce the above 4 commands into 1 and then use 2 parameters for the encoding and collation. this command would look something like:
```
openemr-cmd encoding-collation utf8mb4 utf8mb4-general-ci
```
(this way, the literally hundred or so combinations of encoding and collations are supported with one command)

bradymiller:
Also, lets not embed these scripts in the docker directories since these places are meant for things that go into the dockers during builds. Recommend making another base directory in this repo called `/utilities` and then placing this script and README.md at `/utilities/openemr-cmd/`

After issues are addressed, then this will be ready to bring in :)

docker/openemr/flex-edge/utilities/openemr-cmd/openemr-cmd
bradymiller: btw, if wish to keep shortcut commands like you have for the 3 main calls (would drop the vietnamese one), then that would also make a lot of sense to keep. But should also have a generic command to easily change to the other ones.

reidliu41:
@bradymiller, Actually, I not sure very sure which one in docker that is hardcore or can be a parameter. Please help to recheck, I will modify it, then it will more flexible. ^^
```
 '/root/devtools args'
```

reidliu41:
@bradymiller hi, i got an issue about git which is showed like "utilities/openemr", anything wrong with my operation?

utilities/openemr/openemr-cmd
```diff
@@ -0,0 +1,307 @@
...
+reset_and_reinstall_openemr() {
+	$DOCKER_EXEC_CMD $CONTAINER_ID $SHELL_TYPE '/root/devtools dev-reset-install'
+}
+
+reset_and_reinstall_openemr_with_demodata(){
+	$DOCKER_EXEC_CMD $CONTAINER_ID $SHELL_TYPE '/root/devtools dev-reset-install-demodata'
+}
+
+creat_a_backup_snapshot(){
+	$DOCKER_EXEC_CMD $CONTAINER_ID $SHELL_TYPE '/root/devtools backup example'
```
bradymiller, line 91: oops, didn't notice this. this should also take a parameter rather than hard-code example

utilities/openemr/openemr-cmd
```diff
@@ -0,0 +1,307 @@
...
+}
+
+restore_from_a_snapshot(){
+	$DOCKER_EXEC_CMD $CONTAINER_ID $SHELL_TYPE '/root/devtools restore example'
```
bradymiller, line 95: oops, didn't notice this. this should also take a parameter rather than hard-code example

bradymiller:
@reidliu41 , I just tested it out and really cool stuff. noted 2 more places above where need a parameter. also, move it into the `/utilities/openem-cmd/` directory rather than `/utilities/openemr/` to avoid being to generic (and guessing there will likely be other unrelated `utilities` scripts/tools in the future).

reidliu41:
@bradymiller  replaced the two places you mentioned

bradymiller:
hi @reidliu41 , super cool stuff. bringing this in now. thanks for the contribution!

Been using this already as main way to run the openemr dev docker commands :)
Some things that would be very helpful:
`openemr-cmd --version`
:outputs a version of the script which is defined at top of the script (just nice way to ensure dev has most recent version)
`openemr-cmd up`
:docker-compose up -d
`openemr-cmd down`
:docker-compose down -v
`openemr-cmd stop`
:docker-compose stop
`openemr-cmd start`
:docker-compose start

bradymiller:
two more items on wishlist:
- if the command doesn't match anything, then show the commands/description screen
- ``--help`` shows commands/description screen (this will happen automatically if do above :) )

bradymiller:
Also, feel free to add a link to this script README.md in the https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md (would prob insert it between item 14 and 15 for now)

oh, it just hit me. a command that did the following would also be very nifty (something like `openemr-cmd shell`)
:docker exec -it $(docker ps | grep _openemr | cut -f 1 -d " ") sh
(to quickly open a shell in the docker)

reidliu41:
@bradymiller  can this run like openemr-cmd compose up/down...?  and then I can put it in one function.
```
openemr-cmd up
docker-compose up -d
openemr-cmd down
docker-compose down -v
openemr-cmd stop
docker-compose stop
openemr-cmd start
docker-compose start
```

reidliu41:
@bradymiller, how about item 2, I mean that it can remind the user before they use the command.
```
Also, feel free to add a link to this script README.md in the https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md (would prob insert it between item 14 and 15 for now)
```

reidliu41:
> @bradymiller can this run like openemr-cmd compose up/down...? and then I can put it in one function with a parameter.
> 
> ```
> openemr-cmd up
> docker-compose up -d
> openemr-cmd down
> docker-compose down -v
> openemr-cmd stop
> docker-compose stop
> openemr-cmd start
> docker-compose start
> ```

@bradymiller pls ignore this reply. haha, I know how to do it

bradymiller:
item 2 also works :)
(agree makes much more sense to do as item 2 if we want anyone to know the script exists :) )

PR #259, opened 2020-07-17
Openemr cmd updates

bradymiller:
hi @reidliu41 ,
Did some work on this (after real time testing and use). Basically just made a couple things more flexible (for example can copy the capsule(s) to a specific directory rather than current one). I put your name/email in author on the script (if not ok with it let me know; figured you would be since your name/email is on the commits). Again, really cool stuff, and makes working on the easy dev docker much more efficient (I was previously copy/pasting all those long awkward commands). And definitely let me know if any issues on this PR.

### DIFF
--- a/utilities/openemr-cmd/README.md
+++ b/utilities/openemr-cmd/README.md
@@ -2,22 +2,21 @@
 
 ## Overview
 
-OpenEMR-Cmd is similar to devtools, it helps deveployers to manage and troubleshooting openemr outside the docker,see more detail [here](https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md).
+OpenEMR-Cmd is similar to devtools, it helps developers to manage and troubleshoot openemr outside the docker,see more detail [here](https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md).
 
 ## Implementation
 
 1. Copy the script to local linux environment, create the bin directory if not exist.
 
 ```
-mkdir /home/<username>/bin
-cd /home/<username>/bin
-curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-cmd/openemr-cmd > ./openemr-cmd
+mkdir ~/bin
+curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-cmd/openemr-cmd > ~/bin/openemr-cmd
 ```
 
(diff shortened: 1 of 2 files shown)


reidliu41:
Hi @bradymiller, pretty cool fixed，yeah, it's better. We are the same: more flexible and more efficient. I hope it can help developers operate it easier. Thanks for putting my information in it, I think it's ok.  Please help to merge if you also think ok.
1.  But when I tested, I found something like below:
```
# openemr-cmd unit-test
Running OpenEMR Unit tests
/root/devtools: line 59: vendor/bin/phpunit: not found
# openemr-cmd e2e-test
/root/devtools: line 73: vendor/bin/phpunit: not found
Running OpenEMR e2e tests
```
Is it something change in devtools? Or something wrong with my env?

2. Do we also need to confirm the current what it is? I mean check the current version
```
	if [ $# != 1 ]
	then
		echo 'Please provide the OpenEMR version to upgrade database from.'
		echo 'e.g. openemr-cmd ensure-version 5.0.2'
		exit $BACKUP_FILE_CODE
	else
```

bradymiller:
hi @reidliu41 , guessing the `openemr-cmd docker-log` issue happened for you since the docker was not done setting up. When do an up on the docker it needs to:
1. composer build
2. npm build (gets packages and then builds themes)
3. install openemr
(can take from several minutes to 15 minutes depending on computer power and internet speed; note in places with really slow internet speed can be really painfully slow, which is one of the drivers of all those devtools, since it allows doing a lot of things without needing to do a docker compose down/up)

Regarding the ensure-version command, the underlying devtool is really just meant to test the sql upgrade scripts, and doesn't really need to be the same as the current version (for example, the dev may want to use a different version in the call just to see what happens or breaks)

Bringing this PR in now. @reidliu41 , what you did is really cool stuff, because:
1. the shortcuts makes it easier to both use and understand the underlying devtool commands
2. makes the use of the tools that requires scripting on both the docker and host (those get/put-capsule commands for example) much better and I am guessing those kinds of commands will expand in the future now because of your script
3. an of course it makes the OpenEMR fans are happy:
2DV

reidliu41:
@bradymiller, yep. It seems that not set up. It's working now. I can see the up log in docker-log.
Ok, I understand the ensure-version.
I also agree those commands will expand in the future. haha. Let's look forward to it.

PR #261, opened 2020-07-19
optimize the unecessary functions

reidliu41:
Fixes #  optimize the unnecessary functions
replace the hardcore functions with below:
```
    *)
        $DOCKER_EXEC_CMD $CONTAINER_ID $SHELL_TYPE "$DEV_TOOLS $FIRST_ARG"
        ;;
```

-
-
-

### DIFF
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -35,6 +35,17 @@ quick_open_a_docker_shell(){
 	fi
 }
 
+execute_command_flexible(){
+	if [ $# -lt 3 ]
+	then
+		echo "Please provide the container name and the command."	
+		echo "e.g. `basename $0` [exec|-e] openemr_openemr_1 tail -f /var/log/apache2/error.log"
+	else
+		shift
+		docker exec -it "$@"
+	fi
+}
+
 check_docker_log() {
   docker logs $CONTAINER_ID
 }
@@ -173,6 +184,7 @@ then
   echo "  -h, --help                         Show the commands usage"
   echo "  -v, --version                      Show the openemr-cmd command version"
   echo "  -s, shell                          Open a docker shell quickly"
+  echo "  -e, exec                           Execute commands outside docker"
   echo "  up                                 Execute: docker-compose up -d"
   echo "  down                               Execute: docker-compose down -v"
   echo "  start                              Execute: docker-compose start"
@@ -235,6 +247,9 @@ case "$FIRST_ARG" in
 		;;
 	-s|shell)
 		quick_open_a_docker_shell $2
+		;;
+	-e|exec)
+		execute_command_flexible "$@"
 		;;
 	dl|docker-log)
 		check_docker_log
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -11,6 +11,7 @@ DOCKER_EXEC_CMD='docker exec -i'                                 # docker exec c
 CONTAINER_ID=$(docker ps | grep _openemr | cut -f 1 -d " ")      # get the container id
 SHELL_TYPE='sh -c'                                               # specify shell type
 FIRST_ARG=$1											         # define the first parameter
+DEV_TOOLS=/root/devtools                                         # define the devtools parameter
 
 check_docker_compose_install_or_not(){
 	DOCKER_COMPOSE_CODE=22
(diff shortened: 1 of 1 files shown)


reidliu41:
hi @bradymiller，I delete some functions and combine together, and replace with parameter.

jesdynf:
Hi, Reid.

I don't have any immediate comment -- you and Brady seem to have this well in hand -- but it looks like you're right on the cusp of wanting to make some further structural changes? And on the off-chance you're not familiar with your options, I want to make sure you're exposed to something.

[getops](https://www.shellscript.sh/tips/getopts/) is bash's builtin script parameter handler, allowing you to move from positional to named shell arguments. It can be [leveraged](https://stackoverflow.com/questions/402377/using-getopts-to-process-long-and-short-command-line-options) to support long options as well as single-letter arguments with a little work.

reidliu41:
Hi @jesdynf, thanks, yep, I want to make it simpler and easy to extend. Thank you for your resource. I will check it.

utilities/openemr-cmd/openemr-cmd
```diff
@@ -229,50 +166,51 @@ which docker &>/dev/null
...
+	echo "  st, services-test                  To run services testing"
+	echo "  ft, fixtures-test                  To run fixtures testing"
+	echo "  vt, validators-test                To run validators testing"
+	echo "  ct, controllers-test               To run controllers testing"
+	echo "  ctt, common-test                   To run common testing"
+	echo "  cs, clean-sweep                    To run the entire dev tool suite"
+	echo "  cst, clean-sweep-tests             To run only all the automated tests"
+	echo "  dr, dev-reset                      To reset OpenEMR only"
+	echo "  dri, dev-reset-install             To reset and reinstall OpenEMR"
+	echo "  drid, dev-reset-install-demodata   To reset and reinstall OpenEMR with demo data"
```
bradymiller, line 201: i like the sound if it:
```
drid
```
tenor (4)

kidding aside, though, you keep making devs likes easier. nice that you are working on the insane dev environment. i was planning to update those dockers to fully support devtools (the backup/restore/demo stuff doesn't work), so this is a good reason to do that. when that is up and running, then I would do the following when don't include a docker name:
see if following exist:
```
$(docker ps | grep _openemr_1 | cut -f 1 -d " ")
```
if not there, then see if following exists:
```
$(docker ps | grep _openemr-7-3-312_1 | cut -f 1 -d " ")
```
if not, then use the provided argument

utilities/openemr-cmd/openemr-cmd
```diff
@@ -19,91 +20,43 @@ check_docker_compose_install_or_not(){
...
-check_php_error_logs() {
-	$DOCKER_EXEC_CMD $CONTAINER_ID $SHELL_TYPE '/root/devtools php-log'
+	CONTAINER_NAME=$1
+	CONTAINER_NAME_CODE=24	
+	if [ $# -eq 1 ]
+	then	
+		docker exec -it $CONTAINER_NAME sh
+	else
+		echo '********************************************************'
+		echo 'Access the docker openemr_openemr_1 by default.'
```
bradymiller, line 30: note that this would be more accurate, since the host may not have openemr in a openemr directory:
echo 'Access the docker *_openemr_1 by default.'

utilities/openemr-cmd/openemr-cmd
```diff
@@ -11,6 +11,7 @@ DOCKER_EXEC_CMD='docker exec -i'                                 # docker exec c
 CONTAINER_ID=$(docker ps | grep _openemr | cut -f 1 -d " ")      # get the container id
```
bradymiller, line 11: check out my comment below on what to place in CONTAINER_ID

bradymiller:
Since this is changing commands a bit, rec incrementing the version of the script.

utilities/openemr-cmd/openemr-cmd
reidliu41: @bradymiller, haha, yep. I think openemr-cmd is already almost can cover easy env. But insane env,  I think this "_openemr-7-3-312_1" should gonna be an argument in insane dev.  And the openemr container version, it seems that will keep changing in the future. Not sure it needs another version, or can also cover insane env.

bradymiller: the goal is all the devtools will work on insane environment just like easy dev (just need to do some trickery in the flex docker to support; i think i know what i need to do, hopefully get around to it over the next week or so)

reidliu41:
> Since this is changing commands a bit, rec incrementing the version of the script.

Should be?:
From:
```
openemer-cmd version 1.0.1, build 2020-07-17
```
to:
```
openemer-cmd version 1.0.2, build 2020-07-19
```

utilities/openemr-cmd/openemr-cmd
reidliu41: > note that this would be more accurate, since the host may not have openemr in a openemr directory:
> 
> ```
> echo 'Access the docker *_openemr_1 by default.'
> ```
ok

utilities/openemr-cmd/openemr-cmd
reidliu41: sorry, what is that mean of this one?

bradymiller:
looks good on the version increment

utilities/openemr-cmd/openemr-cmd
bradymiller: I guess I am thinking something like here:
```
INSANE_DEV_DOCKER="_openemr-7-3-312_1"
EASY_DEV_DOCKER="_openemr_1"
CONTAINER_ID=$(docker ps | grep $INSANE_DEV_DOCKER | cut -f 1 -d " ")
if [ -z $CONTAINER_ID ]
then
  CONTAINER_ID=$(docker ps | grep $EASY_DEV_DOCKER | cut -f 1 -d " ")
fi
```
Then the script would work for both easy dev and insane dev.

reidliu41: yep, but seems that not only "_openemr-7-3-312_1" version.  
```
# chkdocker-names openemr
=======Running Docker Names=======
docker_openemr-7-1_1       <<
docker_openemr-7-1-redis_1
docker_openemr-7-2_1          <<
docker_openemr-7-2-redis_1
docker_openemr-7-3-311_1   <<
docker_openemr-7-3-312_1   <<
docker_openemr-7-3-redis-311_1
docker_openemr-7-3-redis-312_1
docker_openemr-edge_1
```

reidliu41: I mean if only need to check _openemr-7-3-312_1 that is ok, if it needs to switch to check multiple versions, seems not work

bradymiller: In the insane development environment, there is one "brain" docker that does all the work (it does the building), which generally uses the most recent alpine version (the easy dev docker also uses this alpine version), which is why there only needs to be one docker to default to if it exists for the insane dev environment. About every 6 months when alpine releases a new version, then would update it on the script.

To support forcing use of other dockers (for example when looking at docker log), then would consider looking into the getops utility posted by @jesdynf, so then command may look something like this if wish to force a docker name (of course, for default behavior would not need it):
```
openemr-cmd -d docker_openemr-7-3-redis-312_1 dl
```
Usage: openemr-cmd [-d docker-name] [ command and command parameters ]

bradymiller:
btw, just let me know when you want me to bring in a PR (to avoid a infinite PR as continue to improve the script :) )

utilities/openemr-cmd/openemr-cmd
reidliu41: Hi @bradymiller, thanks for your explanation. So I will add the default to it.
```
INSANE_DEV_DOCKER="_openemr-7-3-312_1"
EASY_DEV_DOCKER="_openemr_1"
CONTAINER_ID=$(docker ps | grep $INSANE_DEV_DOCKER | cut -f 1 -d " ")
if [ -z $CONTAINER_ID ]
then
  CONTAINER_ID=$(docker ps | grep $EASY_DEV_DOCKER | cut -f 1 -d " ")
fi
```

And also add -d option for support the other dockers. btw, -d idea is really cool. I working on it.

reidliu41:
> btw, just let me know when you want me to bring in a PR (to avoid a infinite PR as continue to improve the script :) )



> btw, just let me know when you want me to bring in a PR (to avoid a infinite PR as continue to improve the script :) )

reidliu41:
> btw, just let me know when you want me to bring in a PR (to avoid an infinite PR as continue to improve the script :) )

Sorry about that. I think it will be ok when the above two functions finished. Hope this week will finish

reidliu41:
Hi @bradymiller, I already add the two functions that you mentioned, I test several times for _openemr and _openemr-7-3-312_1,  and the -d option, I not sure test enough or not. Please help to check if you are available. Thanks. It will cover easy and insane if ok. That would be very cool.

bradymiller:
hi @reidliu41 , Just posted a commit to streamline things a little bit (just let me know if you're ok with it and then will bring it in). All the features tested well on both easy dev, insane dev, and forcing specific dockers with the -d parameter. I also updated the flex dockers so now all the features work in the insane dev.  Just really cool stuff! Can't wait to see what you do next :)

reidliu41:
> hi @reidliu41, Just posted a commit to streamline things a little bit (just let me know if you're ok with it and then will bring it in). All the features tested well on both easy dev, insane dev, and forcing specific dockers with the -d parameter. I also updated the flex dockers so now all the features work in the insane dev. Just really cool stuff! Can't wait to see what you do next :)

hi @bradymiller, it's really cool. Good fixed for shift usage. And the quick shell fixed, I forgot the "-d" option. Great.  I made a quick test for -d that is fine. Cool.
```
shift 3
```
But I got a small question about the s and e option. why was it run without "-"?  Just curious. ^^.
```
  echo "  s, shell                           Open a docker shell quickly"
  echo "  e, exec                            Execute commands outside docker"
```
I think all is ok for me. Please help to merge if you also feel ok.

bradymiller:
hi @reidliu41 , Bringing this in now. Again, thanks for this! Removed the dash in s/shell and e/exec to make them look like standard commands rather than specialty items like -h, -v, and -d (especially -d since that always needs to go first).

PR #266, opened 2020-08-12
Create .gitattributes

boxlady:
Should fix windows to linux/Docker issues.


jesdynf:
I always forget this step until the world ends because of it. Thanks!

boxlady:
NP, I always forget and got tired of change the line endings everytime I
updated when I had to use the laptop ;P


Sincerely,
Julie Buurman

PR #274, opened 2020-10-13
add minikube install

reidliu41:
add minikube install for 
https://github.com/openemr/openemr-devops/tree/master/kubernetes/minikube

Fixes #

-
-
-

### DIFF
--- a/utilities/openemr-env-installer/openemr-env-installer
+++ b/utilities/openemr-env-installer/openemr-env-installer
@@ -420,8 +420,8 @@ startup_the_env() {
 	echo -e "\033[34m==>\033[0mOr run the below command to startup \033[33minsane env\033[0m."
 	echo -e "\033[32m   cd ${code_location}/openemr/contrib/util/docker && openemr-cmd up\033[0m"
 	echo -e "\033[34m==>\033[0mOr run the below command to startup \033[33mOpenEMR Orchestration\033[0m."
-	echo -e "\033[33m   Please run with '--driver=none' for minikube if run as root.\033[0m"
-	echo -e "\033[32m   minikube start --driver=docker\033[0m"
+	echo -e "\033[33m   Please try to run with '--driver=docker' option if it cannot detect the driver.\033[0m"
+	echo -e "\033[32m   minikube start\033[0m"
 	echo -e "\033[32m   cd ${code_location}/openemr-devops/kubernetes/minikube/ && bash kub-up\033[0m"
 	echo '*****************************************************************************************************'
 }
--- a/utilities/openemr-env-installer/openemr-env-installer
+++ b/utilities/openemr-env-installer/openemr-env-installer
@@ -334,7 +334,7 @@ fi
 	then
 		echo -e "\033[30m===>\033[0m \033[32mInstalling minikube... \033[0m"
 		echo ''
-		#curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-${minikube_os}-amd64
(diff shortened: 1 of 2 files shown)


utilities/openemr-env-installer/openemr-env-installer
```diff
@@ -275,12 +300,63 @@ else
...
+then
+	minikube_os=darwin
+fi
+	# Install minikube
+	which minikube &>/dev/null
+	if [ $? -ne 0 ]
+	then
+		echo -e "\033[30m===>\033[0m \033[32mInstalling minikube... \033[0m"
+		echo ''
+		#curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-${minikube_os}-amd64
```
bradymiller, line 337: if this is commented out, then how is it getting the minikube package?

reidliu41: @bradymiller sorry, due to network issue for some test vm, so I comment out for test and forgot to restore

bradymiller:
hi @reidliu41 ,
Really neat stuff. I haven't been able to test this yet and just making sure working on your testing and that the kuberenetes stuff works (or does kubectl also need to be installed?).

reidliu41:
@bradymiller, thanks for your remind, I may need to take some time to test for the Kubernetes startup, and it needs to install kubectl. I will update you if I finished.

reidliu41:
@bradymiller, For the new update, I added kubectl and conntrack that may necessary for minikube.
There are some resource limitation and network issue, I cannot test all distributions. Here are the ubuntu, centos, and Debian for your reference.
https://github.com/reidliu41/test/tree/master/env-installer-test-report/update-minikube-test
During the test, I found it seems it really depends on the VM performance and the network. Basically, it will work if it set up the base services.

utilities/openemr-env-installer/openemr-env-installer
```diff
@@ -382,7 +420,8 @@ startup_the_env() {
 	echo -e "\033[34m==>\033[0mOr run the below command to startup \033[33minsane env\033[0m."
 	echo -e "\033[32m   cd ${code_location}/openemr/contrib/util/docker && openemr-cmd up\033[0m"
 	echo -e "\033[34m==>\033[0mOr run the below command to startup \033[33mOpenEMR Orchestration\033[0m."
-	echo -e "\033[32m   minikube start\033[0m"
+	echo -e "\033[33m   Please run with '--driver=none' for minikube if run as root.\033[0m"
+	echo -e "\033[32m   minikube start --driver=docker\033[0m"
```
bradymiller, line 424: Which driver to use will likely be dependent on the actual environment.
The `--driver=none` means it will install everything on the current os (won't separate it on a vm or docker), so could be problematic for users that do not want that. I didn't even know about the docker driver, which plan to try (I have been using the virtualbox driver).

bradymiller:
hi @reidliu41 , Looks good. Only noted a potential issue with the `--driver=none` since this would make lots of changes to the main os (rather than be abstracted away to a separate vm/docker).

utilities/openemr-env-installer/openemr-env-installer
reidliu41: @bradymiller, yep, because there is some issue about cannot detect the driver. So I will move the none option, and mentioned the docker driver

bradymiller:
hi @reidliu41 , Looking good and bringing this in. Thank for another awesome update and can't wait to see what you do next :)

PR #277, opened 2020-11-24
update raspberry pi documentation

bradymiller:
@jesdynf , cleaned up the raspberry documentation a bit. let me know if ok.


bradymiller:
btw, got 64bit to work (I just needed to use a larger usb flash drive)

jesdynf:
This all looks good! Glad to hear 64-bit stopped biting you.

PR #284, opened 2021-02-21
fix builds for 3.12 series dockers

bradymiller:
fix builds for 3.12 series dockers


bradymiller:
changes also for decreasing size of these dockers
testing well and bringing in

PR #286, opened 2021-02-22
xdebug updates

bradymiller:
fixes: #285

Changed default client port to 9003

Changed the setting to turn on Xdebug to: `XDEBUG_ON` (setting it to 1 turns it on)

Have the following optional settings:
`XDEBUG_IDE_KEY` (not needed in phpstorm or vscode, but maybe another ide will need it)
`XDEBUG_CLIENT_HOST` (should never need, but just in case the xdebug.discover_client_host does not work)
`XDEBUG_CLIENT_PORT` (give flexibility if wish to not use default port 9003)

Testing well in main flex docker on both PHPStorm and VSCode (see this PR in openemr repo for settings that are using to work with this in development environment: https://github.com/openemr/openemr/pull/4248 ) . Profiling is also working.

Gotta bring these changes into 3.13-8 and edge also. And then update the official dockers on dockerhub.

### DIFF
--- a/docker/openemr/flex-3.13-8/autoconfig.sh
+++ b/docker/openemr/flex-3.13-8/autoconfig.sh
@@ -334,31 +334,45 @@ if [ "$REDIS_SERVER" != "" ] &&
     touch /etc/php-redis-configured
 fi
 
-if [ "$XDEBUG_IDE_KEY" != "" ] &&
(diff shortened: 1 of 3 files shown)


bradymiller:
have pushed the new main flex docker to dockerhub.

PR #293, opened 2021-06-17
Alpine 3.14 transition

bradymiller:
Some weirdness is happening on npm build which i remember seeing previously in an older alpine build where there is a permission problem on the `npm install --unsafe-perm` command (it works in the docker build for 6.1.0 but does not work when building in the flex series). As I recall, when i noted this in the older alpine version, it was sorted out on next alpine update (hopefully that will not be the case here and perhaps there is a fix for this that involves more than just waiting around).

Planning to do some more debugging over the next several days.

### DIFF
--- a/docker/openemr/6.1.0/Dockerfile
+++ b/docker/openemr/6.1.0/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.13
+FROM alpine:3.14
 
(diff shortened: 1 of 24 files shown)


bradymiller:
instead completed in this PR: https://github.com/openemr/openemr-devops/pull/299

PR #302, opened 2021-10-16
Future-facing refresh and renewal of Docker scripting

jesdynf:
This PR updates dockers 6.1.0 and flex-edge with the following features:

 * Better Kubernetes support for clustering with discrete administrative and operational roles.
 * Bugfix for non-atomic swarm starts.
 * Minor reorganization of scripts, removing some indirection and breaking large chunks of the main runscripts into secondary stanzas for clarity.
 
Both have been launched and opened with docker-compose; see images jesdynf/openemr:6.1.0 and :flex-edge for further testing.


docker/openemr/6.1.0/Dockerfile
```diff
@@ -54,9 +54,9 @@ ENV APACHE_LOG_DIR=/var/log/apache2
 COPY php.ini /etc/php8/php.ini
 COPY openemr.conf /etc/apache2/conf.d/
 #add runner and auto_configure and prevent auto_configure from being run w/o being enabled
-COPY run_openemr.sh autoconfig.sh auto_configure.php /var/www/localhost/htdocs/openemr/
+COPY openemr.sh ssl.sh xdebug.sh auto_configure.php /var/www/localhost/htdocs/openemr/
```
bradymiller, line 57: wouldn't place the ssl.sh and xdebug.sh scripts in the openemr directory (then can get to the via web browser; prob impossible to run, but wouldn't risk it; especially the xdebug script could be a nasty security vulnerability if bad actor figured out how to run it)

docker/openemr/flex-edge/openemr.sh
```diff
@@ -327,7 +315,7 @@ if [ "$CONFIG" == "1" ] &&
         find . -type f -print0 | xargs -0 chmod 400
 
         echo "Default file permissions and ownership set, allowing writing to specific directories"
-        chmod 700 /var/www/localhost/htdocs/run_openemr.sh
+        chmod 700 /var/www/localhost/htdocs/penemr.sh
```
bradymiller, line 318: typo. missing `o`

bradymiller:
@jesdynf , really nice stuff. When solid (my only issue is placement of those sh scripts in openemr web directory and if it even makes sense to offer xdebug in the production docker line), then should make same changes in `flex-3.13-8` and `flex-3.13` (the gotcha in this one is use of php7 in things like xdebug :) ). Then when in I can do rebuilds on all those dockers (which will especially get quick testing in flex-3.13-8 that feeds the easy dev :) ).

bradymiller:
note I modified above post to point out  flex-3.14 stuff rather than flex-3.13

docker/openemr/6.1.0/Dockerfile
jesdynf: The ssl and xdebug scripts are no more vulnerable than the openemr script itself -- if /that's/ a problem we should fix it, and if it's not we can leave it alone, right?

docker/openemr/flex-edge/openemr.sh
```diff
@@ -340,6 +328,7 @@ if [ "$CONFIG" == "1" ] &&
         rm -f sql_patch.php
         rm -f sql_upgrade.php
         rm -f ippf_upgrade.php
+        rm -f auto_configure.php
```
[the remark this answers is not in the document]
jesdynf: Done.

docker/openemr/6.1.0/Dockerfile
bradymiller: You have a point. But the openemr script is protected by settings (for example, xdebug will not be run unless that setting is on, where is somebody can run xdebug.sh directly then that protection won't work)

bradymiller: One of the reasons i didn't include xdebug in the production line was because i honestly didn't no how secure i could keep it, though.

docker/openemr/flex-edge/openemr.sh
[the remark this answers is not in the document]
jesdynf: There's something odd in that stanza. It runs if `/var/www/localhost/htdocs/auto_configure.php` exists (note path) but it deletes the file found in `/var/www/localhost/htdocs/openemr/` (note path again) before deleting the file `/var/www/localhost/htdocs/auto_configure.php`. This isn't /duplicative/, but is one or the other of these just flat wrong?

docker/openemr/6.1.0/Dockerfile
bradymiller: I mean, it should be impossible to run a bash script from a browser. I think if you simply put following in the xdebug.sh script, then I can sleep better :) :
```
   [ "$XDEBUG_ON" == 1 ]; then
```

docker/openemr/flex-edge/openemr.sh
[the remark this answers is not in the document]
bradymiller: Looks like the production line dockers use /var/www/localhost/htdocs/openemr/auto_configure.php while the development line dockers use /var/www/localhost/htdocs/auto_configure.php which I am pretty sure was done since there is no openemr directory yet during docker building in flex series.
So everything in the script should be using /var/www/localhost/htdocs/auto_configure.php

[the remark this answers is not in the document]
jesdynf: Done.

jesdynf:
3.14 and 3.14-8 updates are in, although untested as yet.

PR #326, opened 2022-08-31
consolidate error logs

### LINKED ISSUE #325: there are 2 separate php error logs
There is now 2 separate php error logs (one for without ssl and one for when using ssl), which makes it more difficult for users to troubleshoot. Is this needed:
https://github.com/openemr/openemr-devops/blob/master/docker/openemr/7.0.1/openemr.conf#L68-L69
(ie. if remove these 2 lines, will the ssl error/access logs then go to the logs defined above that?)

bradymiller:
fixes #325

still testing


bradymiller:
testing well and bringing in (gonna rebuild all the pertinent dockers via the github actions :) )

PR #328, opened 2022-09-27
add quick help tool

reidliu41:
Fixes #

Due to the openemr-cmd -h options became more, so create a qh tool to get the help easily, especially for the newer

-
-
-


bradymiller:
nice to see you back @reidliu41 !!
Perfect timing since just released a intro video on easy dev docker environment that uses your openemr-cmd command (hopefully gonna release more short videos that covers all the things that can be done with openemr-cmd command in future):
https://youtu.be/D4tXP5G9-sY

For the qh command, main concern would be that there could be other competing software/aliases that may be using qh. Would make it openemr specific, maybe something like openemr-cmd-search or openemr-cmd-qh or openemr-cmd-h or something like that.

reidliu41:
@bradymiller wow, that's great.

bradymiller:
@reidliu41 , looks good and bringing in. thanks!

PR #338, opened 2023-01-08
add primary(master)/replica(slave) and persistence to mariadb in kubernetes

bradymiller:
add primary(master)/replica(slave) and persistence to mariadb in kubernetes


bradymiller:
testing well and bringing this in. next up is redis failover :)

PR #342, opened 2023-01-19
Use MYSQL_PORT in setup

### LINKED ISSUE #341: MYSQL_PORT is ignored in setup
### Describe the bug

MYSQL_PORT setting is ignored in setup.

### To Reproduce

Set MYSQL_PORT to some value other than the default 3306 when starting the docker image. Setup tries to connect to MySQL on port 3306 instead of provided MYSQL_PORT.

### Expected behavior

Port provided should be used to connect to MySQL.

### Client configuration

- OpenEMR version: 7.0.1
- Operating system: Alpine (in Docker)

eugeneiiim:
Fixes #341 

MYSQL_PORT setting being ignored

Add `port=` configuration in `prepareVariables` when `MYSQL_PORT` is set.

### DIFF
--- a/docker/openemr/7.0.1/utilities/devtoolsLibrary.source
+++ b/docker/openemr/7.0.1/utilities/devtoolsLibrary.source
@@ -23,6 +23,9 @@ prepareVariables() {
     else
         CUSTOM_DATABASE="openemr"
     fi
+    if [ "$MYSQL_PASS" != "" ]; then
+        CONFIGURATION="${CONFIGURATION} port=${MYSQL_PORT}"
+    fi
     if [ "$OE_USER" != "" ]; then
         CONFIGURATION="${CONFIGURATION} iuser=${OE_USER}"
     fi
--- a/docker/openemr/7.0.1/utilities/devtoolsLibrary.source
+++ b/docker/openemr/7.0.1/utilities/devtoolsLibrary.source
@@ -23,7 +23,7 @@ prepareVariables() {
     else
         CUSTOM_DATABASE="openemr"
     fi
-    if [ "$MYSQL_PASS" != "" ]; then
+    if [ "$MYSQL_PORT" != "" ]; then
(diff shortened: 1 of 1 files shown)


jesdynf:
Oh, goodness, thank you. I can see how we would've missed this but it's nice to have it covered.

This is adjusting the devtools configuration script but not the container setup itself, or any of the docker-compose scripts that launch the container. Is it necessary to review those as well or is the problem you're trying to solve strictly within the devtools?

docker/openemr/7.0.1/utilities/devtoolsLibrary.source
```diff
@@ -23,6 +23,9 @@ prepareVariables() {
     else
         CUSTOM_DATABASE="openemr"
     fi
+    if [ "$MYSQL_PORT" != "" ]; then
+        CONFIGURATION="${CONFIGURATION} port=${MYSQL_PORT}"
+    fi
```
bradymiller, line 28: would rec following how is done in the $MYSQL_DATABASE block above where is converted to $CUSTOM_DATABASE and the default port of 3306 is set if not given. and would then use the $CUSTOM_MYSQL_PORT setting below.

bradymiller:
@eugeneiiim , Thanks for noticing this!
@jesdynf , that first change in the preparevariables function is used on the standard docker openemr configuration, so this will now allowing setting a port in the standard docker install (ie. will be a new setting at `Optional settings for auto installation are:` section in https://hub.docker.com/r/openemr/openemr/ instructions).

docker/openemr/7.0.1/utilities/devtoolsLibrary.source
eugeneiiim: Updated

bradymiller:
hi @eugeneiiim , looks good and bringing in. thanks for the improvements! (fyi, also gonna bring your changes into the other supported dockers, 7.0.0, flex-edge, flex-3.17, flex-3.15-8)

2DV-2(13)

PR #351, opened 2023-02-01
docker fixes to support database ssl/x509 connections

bradymiller:
docker fixes to support database ssl/x509 connections


stephenwaite:
chores

PR #357, opened 2023-05-30
Enable opcache for latest flex and 7.0.1 dockers.

adunsulag:
Enable the opcache module and turn on JIT if xdebug is not enabled.  We benchmarked 50% response time improvements in the API and much faster UX interactions in the regular openemr interface just by having the opcache enabled.  Turning on JIT with xdebug disabled saw another 25% improvement.

So I took a stab at this.  I'm not as familiar with building the docker images but I followed the pattern that we used with the xdebug (Thanks @stephenwaite).

I'd love pointers on what else I should change / update here to make sure I got all of the moving pieces.

### DIFF
--- a/docker/openemr/flex-3.17/opcache.sh
+++ b/docker/openemr/flex-3.17/opcache.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+set -e
+
+if [ ! "$OPCACHE_ON" == 1 ]; then
+   echo bad context for opcache.sh launch
+   exit 1
+fi
+
+if [ ! -f /etc/php-opcache-configured ]; then
+    # install opcache library
+    apk update
+    apk add --no-cache php81-opcache 
+
+    # set up xdebug in php.ini
+    echo "; start opcache configuration" >> /etc/php81/php.ini
+    echo "zend_extension=opcache" >> /etc/php81/php.ini
+    echo "opcache.enable=1" >> /etc/php81/php.ini
+    echo "opcache.enable_cli=0" >> /etc/php81/php.ini
+    echo "opcache.memory_consumption=128" >> /etc/php81/php.ini
+    echo "opcache.max_accelerated_files=100000" >> /etc/php81/php.ini
+    echo "opcache.max_wasted_percentage=5" >> /etc/php81/php.ini
+    echo "opcache.save_comments=1" >> /etc/php81/php.ini
+    if [ ! "$OPCACHE_ON" == 1 ]; then
+	   echo bad context for opcache.sh launch
+	   exit 1
+	fi
+    if [ ! "$XDEBUG_IDE_KEY" != "" ] &&
+	   [ ! "$XDEBUG_ON" == 1 ]; then
+    	   echo "opcache.jit_buffer_size=100M" >> /etc/php81/php.ini
+           echo "opcache.jit=tracing" >> /etc/php81/php.ini
+    fi
+
+    # Ensure only configure this one time
+    touch /etc/php-opcache-configured
+fi
--- a/docker/openemr/flex-edge/opcache.sh
+++ b/docker/openemr/flex-edge/opcache.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+set -e
+
+if [ ! "$OPCACHE_ON" == 1 ]; then
+   echo bad context for opcache.sh launch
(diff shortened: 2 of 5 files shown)


jesdynf (commented):
Good start!

You missed the opcache.sh for the 7.0.1 branch, you need a 7.0.2 branch, and you need to modify openemr.sh to actually run your script (see lines 391-394 of https://github.com/openemr/openemr-devops/blob/master/docker/openemr/7.0.1/openemr.sh ).

docker/openemr/flex-3.17/opcache.sh
```diff
@@ -0,0 +1,36 @@
...
+
+if [ ! "$OPCACHE_ON" == 1 ]; then
+   echo bad context for opcache.sh launch
+   exit 1
+fi
+
+if [ ! -f /etc/php-opcache-configured ]; then
+    # install opcache library
+    apk update
+    apk add --no-cache php81-opcache
```
bradymiller, line 13: 1. would simply install php81-opcache when install all the other packages in Dockerfile?
2. are any of those settings in the default php.ini file? (if so, then could set them there since opcache is always going to be on)?
3. would it make sense to put this in 7.0.2 first (the dev docker that is build nightly from dev) and then do 7.0.1 when sure it is working well since the patch and rebuild of the 7.0.1 docker will be happening very, very soon :) (if testing well on your side, then ok to proceed with 7.0.1 also though)

docker/openemr/flex-edge/opcache.sh
```diff
@@ -0,0 +1,29 @@
...
+    # set up xdebug in php.ini
+    echo "; start opcache configuration" >> /etc/php82/php.ini
+    echo "zend_extension=opcache" >> /etc/php82/php.ini
+    echo "opcache.enable=1" >> /etc/php82/php.ini
+    echo "opcache.enable_cli=0" >> /etc/php82/php.ini
+    echo "opcache.memory_consumption=128" >> /etc/php82/php.ini
+    echo "opcache.max_accelerated_files=100000" >> /etc/php82/php.ini
+    echo "opcache.max_wasted_percentage=5" >> /etc/php82/php.ini
+    echo "opcache.save_comments=1" >> /etc/php82/php.ini
+    echo "opcache.jit_buffer_size=100M" >> /etc/php82/php.ini
```
bradymiller, line 24: this is missing the if ! xdebug thing

docker/openemr/flex-3.17/opcache.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+set -e
+
+if [ ! "$OPCACHE_ON" == 1 ]; then
```
bradymiller, line 5: we will be assuming this is always on, so should not need this block

docker/openemr/flex-3.17/opcache.sh
adunsulag: Yea, I think it makes sense to put it in 7.0.2, was targeting the dev dockers initially and had the thought that we'd want to put it in 7.0.1 but its probably too risky to push this in since we haven't updated the dockers yet.

docker/openemr/flex-3.17/opcache.sh
jesdynf: Honestly the script shouldn't check for xdebug at all, it should Just Run Opcache. The test about xdebug or opcache can be handled in openemr.sh .

bradymiller: i think the issue is that:
1. The docker will always be running opcache no matter what
2. The docker can only run the jit if the xdebug is turned off. And the jit requires new php.ini settings to be inserted.

jesdynf:
So the openemr.sh I think we're after looks like...

```
   [ "$XDEBUG_ON" == 1 ]; then
   sh xdebug.sh
   sh opcache.sh -x
else
  sh opcache.sh
fi
```
And then the opcache script...
```
LOADJIT=1
while getopts 'x' opt; do
  case "$opt" in
    x)
      LOADJIT=0
      ;;
    ?)
      echo -e "Invalid command option.\nUsage: $(basename $0) [-x]"
      exit 1
      ;;
  esac
done
```

And then you can act on LOADJIT's truth or falsity.

bradymiller:
completed in subsequent PR #359

PR #359, opened 2023-07-18
enables opcache on 7.0.2

Jmevorach:
Hello! I saw the pull request for implementing OPcache on the flex and 7.0.1 containers and wanted to give a try at implementing it for 7.0.2.

Would love thoughts and feedback ( @adunsulag @bradymiller @jesdynf and from anyone else interested)!

Hope this message finds you all well.

### DIFF
--- a/docker/openemr/7.0.1/openemr.sh
+++ b/docker/openemr/7.0.1/openemr.sh
@@ -391,6 +391,13 @@ fi
 if [ "$XDEBUG_IDE_KEY" != "" ] ||
    [ "$XDEBUG_ON" == 1 ]; then
    sh xdebug.sh
+else
+   # Configure opcache jit if Xdebug is not being used
+   if [ ! -f /etc/php-opcache-jit-configured ]; then
+      echo "opcache.jit_buffer_size=100M" >> /etc/php82/php.ini
+      echo "opcache.jit=tracing" >> /etc/php82/php.ini
+      touch /etc/php-opcache-jit-configured
+   fi
 fi
 
 if [ "$AUTHORITY" == "yes" ] &&
--- a/docker/openemr/7.0.1/openemr.sh
+++ b/docker/openemr/7.0.1/openemr.sh
@@ -394,8 +394,8 @@ if [ "$XDEBUG_IDE_KEY" != "" ] ||
 else
    # Configure opcache jit if Xdebug is not being used
    if [ ! -f /etc/php-opcache-jit-configured ]; then
-      echo "opcache.jit_buffer_size=100M" >> /etc/php82/php.ini
-      echo "opcache.jit=tracing" >> /etc/php82/php.ini
+      echo "opcache.jit_buffer_size=100M" >> /etc/php81/php.ini
+      echo "opcache.jit=tracing" >> /etc/php81/php.ini
       touch /etc/php-opcache-jit-configured
    fi
 fi
--- a/docker/openemr/7.0.1/openemr.sh
+++ b/docker/openemr/7.0.1/openemr.sh
@@ -391,13 +391,6 @@ fi
 if [ "$XDEBUG_IDE_KEY" != "" ] ||
    [ "$XDEBUG_ON" == 1 ]; then
    sh xdebug.sh
-else
-   # Configure opcache jit if Xdebug is not being used
-   if [ ! -f /etc/php-opcache-jit-configured ]; then
-      echo "opcache.jit_buffer_size=100M" >> /etc/php81/php.ini
-      echo "opcache.jit=tracing" >> /etc/php81/php.ini
-      touch /etc/php-opcache-jit-configured
-   fi
 fi
 
 if [ "$AUTHORITY" == "yes" ] &&
--- a/docker/openemr/7.0.1/php.ini
+++ b/docker/openemr/7.0.1/php.ini
@@ -1764,23 +1764,23 @@ ldap.max_links = -1
 
 [opcache]
 ; Determines if Zend OPCache is enabled
-opcache.enable=1
+;opcache.enable=1
 
 ; Determines if Zend OPCache is enabled for the CLI version of PHP
-opcache.enable_cli=0
+;opcache.enable_cli=0
 
 ; The OPcache shared memory storage size.
-opcache.memory_consumption=128
+;opcache.memory_consumption=128
 
 ; The amount of memory for interned strings in Mbytes.
-opcache.interned_strings_buffer=8
+;opcache.interned_strings_buffer=8
 
 ; The maximum number of keys (scripts) in the OPcache hash table.
 ; Only numbers between 200 and 1000000 are allowed.
-opcache.max_accelerated_files=10000
+;opcache.max_accelerated_files=10000
 
 ; The maximum percentage of "wasted" memory until a restart is scheduled.
-opcache.max_wasted_percentage=5
+;opcache.max_wasted_percentage=5
 
 ; When this directive is enabled, the OPcache appends the current working
 ; directory to the script key, thus eliminating possible collisions between
@@ -1790,12 +1790,12 @@ opcache.max_wasted_percentage=5
 
 ; When disabled, you must reset the OPcache manually or restart the
 ; webserver for changes to the filesystem to take effect.
-opcache.validate_timestamps=1
+;opcache.validate_timestamps=1
 
 ; How often (in seconds) to check file timestamps for changes to the shared
 ; memory storage allocation. ("1" means validate once per second, but only
 ; once per request. "0" means always validate)
-opcache.revalidate_freq=60
+;opcache.revalidate_freq=2
 
 ; Enables or disables file search in include_path optimization
 ;opcache.revalidate_path=0
--- a/docker/openemr/7.0.1/php.ini
+++ b/docker/openemr/7.0.1/php.ini
@@ -1764,23 +1764,23 @@ ldap.max_links = -1
 
 [opcache]
 ; Determines if Zend OPCache is enabled
-;opcache.enable=1
+opcache.enable=1
 
 ; Determines if Zend OPCache is enabled for the CLI version of PHP
-;opcache.enable_cli=0
+opcache.enable_cli=0
 
 ; The OPcache shared memory storage size.
-;opcache.memory_consumption=128
+opcache.memory_consumption=128
 
 ; The amount of memory for interned strings in Mbytes.
-;opcache.interned_strings_buffer=8
+opcache.interned_strings_buffer=8
 
 ; The maximum number of keys (scripts) in the OPcache hash table.
 ; Only numbers between 200 and 1000000 are allowed.
-;opcache.max_accelerated_files=10000
+opcache.max_accelerated_files=10000
 
 ; The maximum percentage of "wasted" memory until a restart is scheduled.
-;opcache.max_wasted_percentage=5
+opcache.max_wasted_percentage=5
 
 ; When this directive is enabled, the OPcache appends the current working
 ; directory to the script key, thus eliminating possible collisions between
@@ -1790,12 +1790,12 @@ ldap.max_links = -1
 
 ; When disabled, you must reset the OPcache manually or restart the
 ; webserver for changes to the filesystem to take effect.
-;opcache.validate_timestamps=1
+opcache.validate_timestamps=1
 
 ; How often (in seconds) to check file timestamps for changes to the shared
 ; memory storage allocation. ("1" means validate once per second, but only
 ; once per request. "0" means always validate)
-;opcache.revalidate_freq=2
+opcache.revalidate_freq=60
 
 ; Enables or disables file search in include_path optimization
 ;opcache.revalidate_path=0
--- a/docker/openemr/7.0.2/openemr.sh
+++ b/docker/openemr/7.0.2/openemr.sh
@@ -393,11 +393,15 @@ fi
 if [ "$XDEBUG_IDE_KEY" != "" ] ||
    [ "$XDEBUG_ON" == 1 ]; then
    sh xdebug.sh
+   #also need to turn off opcache since it can not be turned on with xdebug
+   if [ ! -f /etc/php-opcache-jit-configured ]; then
+      echo "opcache.enable=0" >> /etc/php82/php.ini
+      touch /etc/php-opcache-jit-configured
+   fi
 else
-   # Configure opcache jit if Xdebug is not being used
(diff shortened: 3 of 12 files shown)


docker/openemr/7.0.2/openemr.sh
jesdynf, line 1: Almost right -- this script is run every time the container is launched, and you'll append these directives every time the container's launched, so at the minimum you need to use the touch/if pattern here to make sure it's only done once.

I wonder, though, if making use of https://stackoverflow.com/questions/1391808/how-do-i-include-a-php-ini-file-in-another-php-ini-file wouldn't be wiser -- if Alpine supports it, and I think it does, you could have a php.ini fragment here and then depending on if XDEBUG was set or not, you could turn JIT on or off to allow you to cleanly modify the container and relaunch.

Jmevorach: That's great feedback; thanks! I added something for the touch/if pattern.

I definitely like the idea of separate ini files and then loading them conditionally. I think it may be easier for others to look through an ini file and see what configurations were being made rather than reading through a script. 

If it's alright with you I'd like to open another PR for that only because as long as we're moving the OPcache JIT configuration into a separate ini file I'd like to try also moving the xdebug php.ini configuration done in xdebug.sh into its own ini file as well. 

The end result would be something like this:

`if [ "$XDEBUG_IDE_KEY" != "" ] ||
   [ "$XDEBUG_ON" == 1 ]; then
   sh xdebug.sh
else
   # Configure opcache jit if Xdebug is not being used
   <code that configures php to use an additional ini file with the OPCache JIT configuration>
fi`

and inside xdebug.sh lines 11-49 would be replaced with code that configures php to use an additional ini file with the xdebug configuration.

Jmevorach: That's great feedback; thanks! I added something for the touch/if pattern.

I definitely like the idea of separate ini files and then loading them conditionally. I think it may be easier for others to look through an ini file and see what configurations were being made rather than reading through a script.

If it's alright with you I'd like to open another PR for that only because as long as we're moving the OPcache JIT configuration into a separate ini file I'd like to try also moving the xdebug php.ini configuration done in xdebug.sh into its own ini file as well.

The end result would be something like this:

```
   [ "$XDEBUG_ON" == 1 ]; then
   sh xdebug.sh
else
   # Configure opcache jit if Xdebug is not being used
   <code that configures php to use an additional ini file with the OPCache JIT configuration>
fi
```

and inside xdebug.sh lines 11-49 would be replaced with code that configures php to use an additional ini file with the xdebug configuration.

Jmevorach:
Hello! I pushed a commit that implements OPCache for 7.0.1, flex-3.17 and flex-3.18. I've tested 7.0.2 and it works (tested on an ARM cpu) in the sense that I was able to start OpenEMR and poke around the console for a few minutes. I think the implementation should probably work for the others as well (especially for flex-3.18 which also uses php-82 rather than php-81) but having said that I've only tested for 7.0.2. 

If it's alright with you I think I'd like to try to merge the current work before proceeding to any refactoring in regards to additional ini files (of course only when the code is thought to be ready to merge and I'm more than happy to do more work if there's things I should do to help the code reach that state). 

My thinking is that if we implement this as a few smaller PRs rather than one large one that we will be more likely to catch any bugs in testing and we'll have an easier time reverting smaller changes without having to roll back everything in the event a bug manages to make it through testing. 

Having said that I'll of course take your lead here and am happy to proceed in any way you'd like me to. The way I see it we can either ... 

- Try to merge the current work and then do smaller PRs for implementing separate ini files for both Xdebug and also OPcache JIT. 
- Iterate on this PR to have it include all the changes involved in implementing separate ini files for both Xdebug and also OPcache JIT.
- Close this PR and open another PR with a different name where I have it include all the changes involved in implementing separate ini files for both Xdebug and also OPcache JIT in addition to the current work done to implement OPcache.
- Do something else you'd prefer not mentioned above.

Totally up to you though. I trust your judgement in how we should proceed here. However we implement this I can say that at least for me the user interface seems _significantly_ faster with OPcache turned on so it's going to be fun and exciting for me to work on no matter how we decide to do things. 

Also I was thinking if the idea with the separate ini files is to make it so that we can do something like rerun openemr.sh with different settings to toggle the configuration couldn't we potentially do something like leave everything else as is but have openemr.sh at the top of the script make a copy of the original php.ini if it doesn't already exist and then for every subsequent rerun start by restoring the php.ini to its original state so that any subsequent configuration alteration reflect the currently requested configuration unaltered by any previously requested configurations? Just a thought but I'd be curious to know what you thought about it.

Hope this message finds you well!

Jmevorach:
Hello! I pushed a commit that implements OPCache for 7.0.1, flex-3.17 and flex-3.18. I've tested 7.0.2 and it works (tested on an ARM CPU) in the sense that I was able to start OpenEMR and poke around the UI for a few minutes. I think the implementation should probably work for the others as well (especially for flex-3.18 which also uses php-82 rather than php-81) but having said that I've only tested for 7.0.2.

If it's alright with you I think I'd like to try to merge the current work before proceeding to any refactoring in regards to additional ini files (of course only when the code is thought to be ready to merge and I'm more than happy to do more work if there are more things I should do to help the code reach that state).

My thinking is that if we implement this as a few smaller PRs rather than one large one that we will be more likely to catch any bugs in testing and we'll have an easier time reverting smaller changes without having to roll back everything in the event a bug manages to make it through testing.

Having said that I'll of course take your lead here and am happy to proceed in any way you'd like me to. The way I see it we can either ...

- Try to merge the current work and then do smaller PRs for implementing separate ini files for both Xdebug and also OPcache JIT.
- Iterate on this PR to have it include all the changes involved in implementing separate ini files for both Xdebug and also OPcache JIT.
- Close this PR and open another PR with a different name where I have it include all the changes involved in implementing separate ini files for both Xdebug and also OPcache JIT in addition to the current work done to implement OPcache.
- Do something else you'd prefer not mentioned above.

Totally up to you though. I trust your judgement in how we should proceed here. However we implement this I can say that at least for me the user interface seems significantly faster with OPcache turned on so it's going to be fun and exciting for me to work on no matter how we decide to do things.

Also I was thinking if the idea with the separate ini files is to make it so that we can do something like rerun openemr.sh with different settings to toggle the configuration couldn't we potentially do something like leave everything else as is but have openemr.sh at the top of the script make a copy of the original php.ini if it doesn't already exist and then for every subsequent rerun start by restoring the php.ini to its original state so that any subsequent configuration alteration reflect the currently requested configuration unaffected by any previously requested configurations? Just an idea but I'd be curious to know what you thought about it.

Hope this message finds you well!

jesdynf:
Looks good to me!

@bradymiller @adunsulag Can either of you glance at these opcache settings, or comment on how they might interact with the particulars of flex? I can't do that -- I don't really, uhm, PHP any -- but we're good on structure and the issues from the other PR are largely covered.

docker/openemr/7.0.1/php.ini
```diff
@@ -1790,12 +1790,12 @@ ldap.max_links = -1
...
 ; When disabled, you must reset the OPcache manually or restart the
 ; webserver for changes to the filesystem to take effect.
-;opcache.validate_timestamps=1
+opcache.validate_timestamps=1
 
 ; How often (in seconds) to check file timestamps for changes to the shared
 ; memory storage allocation. ("1" means validate once per second, but only
 ; once per request. "0" means always validate)
-;opcache.revalidate_freq=2
+opcache.revalidate_freq=60
```
bradymiller, line 1798: Looks like this is the only setting that will be different from default (the other settings above are not necessary since they are already set to that by default: https://www.php.net/manual/en/opcache.configuration.php)

docker/openemr/7.0.1/openemr.sh
```diff
@@ -391,6 +391,13 @@ fi
 if [ "$XDEBUG_IDE_KEY" != "" ] ||
    [ "$XDEBUG_ON" == 1 ]; then
    sh xdebug.sh
+else
+   # Configure opcache jit if Xdebug is not being used
+   if [ ! -f /etc/php-opcache-jit-configured ]; then
+      echo "opcache.jit_buffer_size=100M" >> /etc/php81/php.ini
+      echo "opcache.jit=tracing" >> /etc/php81/php.ini
```
bradymiller, line 398: the line above not needed since this is the default setting: https://www.php.net/manual/en/opcache.configuration.php
(note that the opcache.jit_buffer_size set to 100M is needed though since is not default)

jesdynf: Hang on, if it's not needed because it's default then it's *default on* which means that the real problem is that's *not off* on the other branch and we need to fix that instead.

docker/openemr/7.0.1/php.ini
bradymiller: Actually I would just keep this setting at default. @adunsulag was having a very good performance improvement with this setting on default of 2. The problem with 60 is that it will kill the developers (ie. imagine needing to wait 60 seconds for a code change to be implemented while developing/testing :) ).

docker/openemr/7.0.1/openemr.sh
```diff
@@ -391,6 +391,13 @@ fi
...
+      touch /etc/php-opcache-jit-configured
+   fi
 fi
```
bradymiller, line 401: Recommend doing the following in above if/else clause (and then no changes are needed in the php.ini file):
```
   [ "$XDEBUG_ON" == 1 ]; then
   sh xdebug.sh
   #also need to turn off opcache since it can not be turned on with xdebug
   if [ ! -f /etc/php-opcache-jit-configured ]; then
      echo "opcache.enable=0" >> /etc/php81/php.ini
      touch /etc/php-opcache-jit-configured
   fi
else
   # Configure opcache jit if Xdebug is not being used (note opcache is already on, so just need to add setting(s) to php.ini that are different from the default setting(s))
   if [ ! -f /etc/php-opcache-jit-configured ]; then
      echo "opcache.jit_buffer_size=100M" >> /etc/php81/php.ini
      touch /etc/php-opcache-jit-configured
   fi
fi
```

docker/openemr/7.0.1/php.ini
jesdynf: We could do php.ini.prod and php.ini.flex and then pick one or the other -- one with a sixty-second clock and one with a two-second clock? Pass in arguments to the build to pick the one or the other. Might save it for later though.

docker/openemr/7.0.1/openemr.sh
jesdynf: Just confirming, it's okay to set php.ini variables multiple times and the last one governs? No need to replace the existing setting in the file?

docker/openemr/7.0.1/php.ini
bradymiller: Note not too attached to what that setting is since the standard dev environments have xdebug going, so this won't be on anyways. However, rather than changing here, would just change it in the if/else block I noted below.

bradymiller:
Would only make the changes in the following dockers (would not do 7.0.1 since that will be used to build a production version in couple weeks when the next 7.0.1 patch is released; ie. lets ensure testing well for awhile before going to 7.0.1):
7.0.2
flex-3.18
flex-edge

bradymiller:
hi @Jmevorach , just completed a review. thanks for taking this on! Looking forward to see where things go.

docker/openemr/7.0.1/openemr.sh
bradymiller: yep, ended up doing that in the if/else in comment below (turns it off when xdebug is on :) )

docker/openemr/7.0.1/openemr.sh
bradymiller: yes, that is my understanding. note we don't even need to touch php.ini since the setting changes are minimal (note could do the `opcache.revalidate_freq=60` here if wish also).

bradymiller:
actually, this gets a bit more interesting :)
Looks like xdebug will work with opcache if opcache is loaded first:
https://www.php.net/manual/en/opcache.installation.php

That being said, isn't this PR missing the following setting when adding things to the php.ini:
`zend_extension=opcache`

If that is the case, then if we simply run the opcache config stuff before xdebug config stuff, then it should work with xdebug (in theory of course :) ).

Jmevorach:
Hello all! Thanks for the review @bradymiller! Really appreciate it. And no problem this is a lot of fun to work on :) . 

I made some updates to the PR to align with the feedback. If I missed the mark on anything please let me know.

I've removed the changes to 7.0.1 and I've made updates to the changes to 7.0.2, flex-3.17, and flex-3.18. 

**Couple of things:**

- I left commented out `zend_extension=opcache` because if I commented it it back in or even pointed to the absolute path of opcache.so I got a message back when I went to run the container saying `"Cannot load Zend OPcache - it was already loaded"`. My theory (although I haven't confirmed this) is that in order to cut down on package size the php82 package for Alpine doesn't include the OPcache extension by default. It also seems that installing `php82-opcache` results in OPcache being loaded into php82. 
- Xdebug has a compatibility section here with some good documentation on OPcache and OPcache JIT (https://xdebug.org/docs/compat#compat). The short of it is is that Xdebug can be loaded together with OPcache if you load OPcache first but it's not `"100% compatible"` and `"When Opcache's optimiser is enabled, some functionality in Xdebug will not work, or differently, because it removes or changes PHP's VM instructions that Xdebug relies on"` and `"Xdebug does not work with PHP's JIT engine, which is implemented in Opcache. When Xdebug is loaded with PHP's JIT on, you will get a warning, and JIT will be disabled."`. So it looks like things _might_ work if we just load OPCache and Xdebug without JIT enabled but I haven't tested that.

As before I tested the current PR for 7.0.2 and it works (tested on an ARM CPU) in the sense that I was able to start OpenEMR and poke around the UI for a few minutes. 

Hope this message finds everyone well!

bradymiller:
nice work @Jmevorach !!!! (agree to not combine xdebug and opcache :) )

dance-bear

Jmevorach:
Got you! Thanks and thank you all for your help with this! Was a lot of fun :) .

adunsulag:
@Jmevorach thanks for driving this to the finish line.  I've been bogged down with other stuff so I'm happy to see you pick this up and complete it.

Jmevorach:
No problem @adunsulag happy to help! Really happy to be a contributor to an OpenEMR repository :) .

Jmevorach:
And thanks to everyone for the kudos; really appreciate it :) .

stephenwaite:
jumping

PR #361, opened 2023-07-24
Adds php82-pecl-apcu to Dockerfile, adds APCu optimizations to autolo…

Jmevorach:
…ader, uses OPcache for auto_configure.php and uses multiple processes for make and find.

- Adds php82-pecl-apcu package which allows us to do level 2/B optimizations with the autoloader as described here: https://getcomposer.org/doc/articles/autoloader-optimization.md#optimization-level-2-b-apcu-cache
- Leverages the OPcache file cache to run auto_configure.php faster. 
- Uses multiple processes for Make and find which was found to reduce runtime there as well.

Tested for 7.0.2!

### DIFF
--- a/docker/openemr/7.0.2/Dockerfile
+++ b/docker/openemr/7.0.2/Dockerfile
@@ -24,7 +24,7 @@ RUN apk add --no-cache git build-base libffi-dev python3-dev cargo \
     && git clone https://github.com/openemr/openemr.git --depth 1 \
     && rm -rf openemr/.git \
     && cd openemr \
-    && composer install --no-dev --optimize-autoloader --apcu-autoloader \
+    && composer install --no-dev \
     && npm install --unsafe-perm \
     && npm run build \
     && cd ccdaservice \
--- a/docker/openemr/7.0.2/Dockerfile
+++ b/docker/openemr/7.0.2/Dockerfile
@@ -7,7 +7,7 @@ RUN apk add --no-cache \
     php82-json php82-pdo php82-pdo_mysql php82-curl php82-ldap php82-openssl php82-iconv \
     php82-xml php82-xsl php82-gd php82-zip php82-soap php82-mbstring php82-zlib \
     php82-mysqli php82-sockets php82-xmlreader php82-redis php82-simplexml php82-xmlwriter php82-phar php82-fileinfo \
-    php82-sodium php82-calendar php82-intl php82-opcache \
+    php82-sodium php82-calendar php82-intl php82-opcache php82-pecl-apcu \
     perl mysql-client tar curl imagemagick nodejs npm \
     python3 openssl py-pip openssl-dev dcron \
     rsync shadow ncurses \
@@ -24,7 +24,7 @@ RUN apk add --no-cache git build-base libffi-dev python3-dev cargo \
     && git clone https://github.com/openemr/openemr.git --depth 1 \
     && rm -rf openemr/.git \
     && cd openemr \
-    && composer install --no-dev \
+    && composer install --no-dev --optimize-autoloader --apcu-autoloader \
     && npm install --unsafe-perm \
     && npm run build \
     && cd ccdaservice \
@@ -34,7 +34,7 @@ RUN apk add --no-cache git build-base libffi-dev python3-dev cargo \
     && /root/.composer/vendor/bin/phing vendor-clean \
     && /root/.composer/vendor/bin/phing assets-clean \
     && composer global remove phing/phing \
-    && composer dump-autoload -o \
+    && composer dump-autoload --optimize --apcu \
     && composer clearcache \
     && npm cache clear --force \
     && rm -fr node_modules \
--- a/docker/openemr/7.0.2/openemr.sh
+++ b/docker/openemr/7.0.2/openemr.sh
@@ -24,6 +24,8 @@ swarm_wait() {
 auto_setup() {
     prepareVariables
 
+    chmod -R 600 .
+
     #create temporary file cache directory for auto_configure.php to use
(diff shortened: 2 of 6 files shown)


docker/openemr/7.0.2/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN apk add --no-cache git build-base libffi-dev python3-dev cargo \
     && git clone https://github.com/openemr/openemr.git --depth 1 \
     && rm -rf openemr/.git \
     && cd openemr \
-    && composer install --no-dev \
+    && composer install --no-dev --optimize-autoloader --apcu-autoloader \
```
bradymiller, line 27: really cool stuff! nice read on it here:
https://getcomposer.org/doc/articles/autoloader-optimization.md

Looks like above command parameter and the below dump-autoload command are redundant. It only needs to be in the dump-autoload below, so would keep the parameters there, but no reason to have in above line (just in case for some reason we need to change/modify them int he future, best to have them in only one place).

docker/openemr/7.0.2/openemr.sh
```diff
@@ -24,8 +24,26 @@ swarm_wait() {
...
+    echo "opcache.file_cache_consistency_checks=1" >> auto_configure.ini
+    echo "opcache.enable_file_override=1" >> auto_configure.ini
+    echo "opcache.max_accelerated_files=1000000" >> auto_configure.ini
+
+    #run auto_configure
+    php auto_configure.php -c auto_configure.ini -f ${CONFIGURATION} || return 1
+
+    #remove temporary file cache directory and auto_configure.ini
+    rm -r $TMP_FILE_CACHE_LOCATION
+    rm auto_configure.ini
```
bradymiller, line 46: also neat stuff! (very curious to see how much faster the config happens)

docker/openemr/7.0.2/openemr.sh
```diff
@@ -322,10 +340,14 @@ if
             # This section only runs once after per docker since auto_configure.php gets removed after this script
 
             echo "Setting user 'www' as owner of openemr/ and setting file/dir permissions to 400/500"
+
+            #return number from nproc to have value for -P flag in xargs
+            N_PROC=$(nproc --all)
```
bradymiller, line 345: another neat optimization!

bradymiller:
hi @Jmevorach , very nice stuff! Review complete. Only thing that noted is likely no need for redundant parameters on the `composer install` line (since the parameters are needed in the `composer dump-autoload` line). If addressed and testing well for you, then would bring your changes also into the `flex-edge` and `flex-3.18` dockers. Then would be ready to bring this PR in.

jesdynf:
Are we sure this belongs in flex at all? My reading of the docs indicated
this was a production-only tool that actively interfered with development.
Did I misunderstand something?

bradymiller:
The level 1 class optimization for composer is already in the dockers (and flex) via the `dump-autoload -o`. And the 2b (acpu stuff) shouldn't break dev stuff unless i am missing something. (since they will still do psr4 seek on drive for a class if it is missing; from my read on how it works; i could be wrong on that :) ). Note that flex is not only used for development; a advanced user can also use it as a way to serve a production instance from a customized git repo/branch. If the feature will not work for development, though, then may need to use a setting (perhaps use the already existent `DEVELOPER_TOOLS: "yes"`) to skip it if on.

docker/openemr/7.0.2/Dockerfile
Jmevorach: Made the change! Thanks for pointing that out!

docker/openemr/7.0.2/openemr.sh
Jmevorach: Thanks!

docker/openemr/7.0.2/openemr.sh
Jmevorach: Thanks! It's a really considerable speed increase. Around ~%40 runtime reduction although I'd be really curious to see what others observe.

Jmevorach:
Thanks all for the review! I think the APCu autoloader optimization feature should probably work for development. I did some research and from what I found `--classmap-authoritative` is the fastest optimization you can apply to the autoloader but as stated in the documentation linked to above `"it also means that in case a class is generated at runtime for some reason, it will not be allowed to be autoloaded. If your project or any of your dependencies does that then you might experience "class not found" issues in production. Enable this with care."`. 

The `--apcu-autoloader` optimization makes the autoloader faster and the only trade off is that it requires the use of APCu memory and requires that you have the required packages to do so which in our case just required the addition of the `php82-pecl-apcu` package to the `apk add` command in the Dockerfile. It should be safe to use in development as well but we'll always need to make sure that the appropriate APCu package(s) are installed so as to make that possible.

I committed additional changes to flex-3.18 and flex-edge and fixed my original commit for 7.0.2.. I tested on 7.0.2! Hope this message finds everyone well. This was a lot of fun to work on :) .

bradymiller (approved):

docker/openemr/7.0.2/openemr.sh
```diff
@@ -24,8 +24,26 @@ swarm_wait() {
 auto_setup() {
     prepareVariables
 
-    chmod -R 600 .
```
bradymiller, line 27: hi @Jmevorach ,
Was about to bring this in, but noted the above line was removed. Was removing above line needed for below changes to work?

Jmevorach: Hello @bradymiller! Good catch; apologies for that that was a mistake. Didn't mean to take that out. I put it back in and tested again for 7.0.2!

PR #365, opened 2023-08-09
feat: support javascript unit test in devtools and openemr-cmd

bradymiller:
support javascript unit test in devtools and openemr-cmd


bradymiller:
testing well

PR #379, opened 2023-12-27
migrate 7.0.3-dev docker to alpine 3.19 with php 8.3

bradymiller:
migrate 7.0.3-dev docker to alpine 3.19 with php 8.3


stephenwaite:
excited_elephant

PR #382, opened 2024-02-12
Specify current docker image

surfacehazard:
Specify current docker image used so it can be overridden.

Fixes #

-
-
-

### DIFF
--- a/packages/lightsail/launch.sh
+++ b/packages/lightsail/launch.sh
@@ -13,7 +13,7 @@ exec > /var/log/lightsail-launch.log 2>&1
 SWAPAMT=1
 SWAPPATHNAME=/mnt/auto.swap
 
-CURRENTDOCKER=openemr:latest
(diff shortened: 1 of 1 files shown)


jesdynf:
I'm not sure I understand the issue. The `-t` option accepts an override image, which gets sed-replaced into the file if necessary during .yml selection in lines 81-111, and I don't know why `7.0.2` would work if `latest` doesn't. Is there a problem with that logic I'm not aware of?

surfacehazard:
sed won't replace anything - there is no openemr:latest in the docker-compose files

jesdynf:
:tantrum follows:

jesdynf:
I set this to :latest so I wouldn't have to keep catching it up, and I can't *just* change the dependencies to latest because some of them, the dev ones, actually honestly do have to hardcode 7.0.2, so you're right, this was the only option. Thanks for catching this!

PR #391, opened 2024-05-09
speed up chmod

aapjeisbaas:
This will speed up chmod by only executing chmod on non conforming items This reduces disk iops, and improves start times.

#### Short description of what this resolves
- slow startup/scaleup/redeploy while waiting for chmod

#### Changes proposed in this pull request
- only chmod files and dirs that need changing

#### Extra notes
I excluded `'./openemr.sh'` to prevent it being changed to 400

I excluded `"sites/default/documents/smarty/main/*"` because I'm not sure if the chmod is needed here as thought this might fill up with a lot of generated files and folders which might slow the process down.

#### Questions
- The files and folders in path `sites/default/documents` are set to `700` was this intentional?

### DIFF
--- a/docker/openemr/7.0.3/openemr.sh
+++ b/docker/openemr/7.0.3/openemr.sh
@@ -24,7 +24,7 @@ swarm_wait() {
 auto_setup() {
     prepareVariables
 
-    chmod -R 600 .
+    find . -not -perm 600 -exec chmod 600 {} \+
 
     #create temporary file cache directory for auto_configure.php to use
     TMP_FILE_CACHE_LOCATION="/tmp/php-file-cache"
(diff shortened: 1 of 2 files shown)


jesdynf:
Thanks for this! I really appreciate you taking the time to submit it.

Do me one favor though -- add these to the flex-3.19 branch too, please? Just that one and no others though.

aapjeisbaas:
@jesdynf I've added the same type of chmod to the 3.19 docker 
If someone can further validate this that would be great.
I will do some local testing but more independent checks are welcome.

bradymiller:
nice work @aapjeisbaas ! Sorry for the delayed review on this. Haven't been able to work on kubernetes stuff in awhile, but those original permissions changes were very slow. Looking forward to seeing the performance improvement of your new changes. Also brought in your changes to the flex-edge docker at https://github.com/openemr/openemr-devops/pull/392

PR #395, opened 2024-08-07
update openemr-cmd to be flexible for - or _ in container names

bradymiller:
update openemr-cmd to be flexible for - or _ in container names


bradymiller:
@stephenwaite , testing well on non-snap docker. Very confident will work well on snap docker (since same grep strategy is working on the github actions ci where using dash). Let me know though if it isn't working.

PR #406, opened 2025-02-10
change deprecated mysql commands to mariadb commands

bradymiller:
change deprecated mysql commands to mariadb commands


stephenwaite:
sheep

PR #407, opened 2025-02-10
fix random patient generator

bradymiller:
fix random patient generator


bradymiller:
Will also need to update the dockers after bring this in

Issue #396, opened 2024-08-08
Setup script failing on mac


ElijahBus:
### Describe the bug

When running the script `openemr-cmd up` on mac , the script is checking for `docker-compose` to be installed. while this is applicable to most linux distributions, docker desktop app on mac gives the possibility of executing `docker-compose` with `docker compose ...` ( no hyphen ).

Not sure whether this should be a bug or a feature request. because it is still possible to install `docker-compose` on mac, but why ? while we are already offered the same capabilities with `docker compose`

### To Reproduce

1. Install docker desktop on mac
2. Setup openemr & openemr cmd
3. `cd` openemr/docker/development-easy-arm
4. Run : openemr-cmd 
5. This error is logged:
<img width="672" alt="image" src="https://github.com/user-attachments/assets/b85d82bb-deb4-43a7-914e-db87a8e321b1">

### Expected behavior

It should set up the local development considering that docker is already installed

### Client configuration

- OpenEMR version: 7
- Operating system: MacOs Sonoma

camrionnvmff:
The issue you encountered is caused by differences in command format when using the docker compose command on macOS. The Docker Desktop application on macOS has integrated Docker compose into Docker commands in the new version and no longer requires separate installation of Docker compose commands. Therefore, the new command format is docker compose (without hyphens).
In fact, Docker and Mac systems are not compatible. If you are using Mac, perhaps you can try using the Servbay tool to solve your needs.

bradymiller:
Hi @ElijahBus , Would consider this a bug fix. Now it makes sense why the github actions ci engine broke recently (docker-compose no longer existed so needed to change it to docker compose). Looks like docker-compose (V1) is getting deprecated for docker compose (V2).
https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose
https://docs.docker.com/compose/migrate/

I think to fix this, just need to be more flexible at https://github.com/openemr/openemr-devops/blob/master/utilities/openemr-cmd/openemr-cmd#L33-L37
(if the docker-compose check fails, then need to check for docker compose)
(also need to then set the working command as a variable, so it is used correctly in the 4 places where the scripts uses that command)

ElijahBus:
[...]

PR #415, opened 2025-04-24
mws/shellcheck openemr cmd

msummers42:
This PR addresses all issues in `openemr-cmd` detected by [Shellcheck, a shell script static analyzer](https://github.com/koalaman/shellcheck) and begins to align the script's conventions and style with [Google's shell guide](https://google.github.io/styleguide/shellguide.html). This should aid in maintainability as well as preventing defects.

It's easy and useful to debug this script using `bash -x ./openemr-cmd` in the `utilities/openemr-cmd` directory.

Beyond a purely cosmetic fixes for some typos & indentation, this PR adds "strict mode", consistent variable quoting, standardizes the use of the bracket test construct. I've added a couple of functions to help clean up & centralize the in-container invocation of the devtools script. I saw an issue discussing moving that somewhere else. If that change occurs, it will be easy to support forwards/backwards compatibility by adding a little logic around the declaration of the  `DEV_TOOLS` variable.

There is a lot of opportunity to further refactor,  so I'm thinking of this as a first pass.

Also added function to treat the issue of backwards/forwards compatibility with docker compose v2, issue here https://github.com/openemr/openemr-devops/issues/396

### DIFF
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -477,14 +477,15 @@ fi
 # override the default container setting if supplied via the -d parameter
 #  (also removing no longer needed parameters via shift)
 CHANGE_CONTAINER_CODE=25
-if [[ "${FIRST_ARG}" == '-d' ]] && [[ $# -ge 3 ]]; then
-	CONTAINER_ID="$2"
-	FIRST_ARG="$3"
-	shift 3
-elif [[ "${FIRST_ARG}" == '-d' ]] && [[ $# -lt 3 ]]; then
-	echo 'Please provide a docker id/name when using -d parameter.'
-	echo 'e.g. openemr-cmd -d docker_openemr-7-3-redis-312_1 dl'
-	exit "${CHANGE_CONTAINER_CODE}"
+if [[ "${FIRST_ARG}" = '-d' ]]; then
+  if (( $# < 3 )); then
+    echo 'Please provide a docker id/name when using -d parameter.'
+    echo 'e.g. openemr-cmd -d docker_openemr-7-3-redis-312_1 dl'
+    exit "${CHANGE_CONTAINER_CODE}"
+  fi
+  CONTAINER_ID="$2"
+  FIRST_ARG="$3"
+  shift 3
 else
   shift
 fi
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -355,8 +355,8 @@ fi
 # Script usage.
 if [[ $# -eq 0 || "${FIRST_ARG}" = '--help' || "${FIRST_ARG}" = '-h' ]]; then
   echo ""
-  echo "Usage: $(basename "$0") COMMAND [ARGS]"
-  echo "Usage: $(basename "$0") -d <docker name> COMMAND [ARGS]"
+  echo "Usage: ${0##*/} COMMAND [ARGS]"
+  echo "Usage: ${0##*/} -d <docker name> COMMAND [ARGS]"
   echo "Options:"
   echo "  -h, --help                         Show the commands usage"
   echo "  -v, --version                      Show the openemr-cmd command version"
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -8,13 +8,15 @@ set -euo pipefail
 # Version   : See below --version section
 # Copyright : Copyright (c) 2020 Reid Liu <guliu@redhat.com>
 # Copyright : Copyright (c) 2020-2025 Brady Miller <brady.g.miller@gmail.com>
+# Copyright : Copyright (c) 2025 Matt Summers <matt@openCoreEMR.com> 
 # Author    : Reid Liu <guliu@redhat.com>
 # Author    : Brady Miller <brady.g.miller@gmail.com>
+# Author    : Matt Summers <matt@openCoreEMR.com>
 # License   : https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
 #################################################################################################
 
 # Increment the version when modify script
-VERSION="1.0.19"
+VERSION="1.0.20"
 
 # If the docker is snap or non-snap docker
 # Setting the container names accordingly
@@ -83,13 +85,14 @@ check_docker_log() {
 
 check_docker_names() {
 	if [[ $# -eq 1 ]]; then
-        local CHECK_NAME="$1"     # specify the checking name
 		# ARGS: couchdb php fhir mariadb mysql nginx openemr openldap orthanc redis.
+        # specify the name to check
+        local CHECK_NAME="$1"
 		echo '===Check the single Docker Name==='
 		echo '  STATUS        NAME'
         # Capture output, check if empty
         local matching_container
-        matching_container=$(docker ps -a --format "{{.Status}}\t{{.Names}}"| grep "${CHECK_NAME}" || true) # Keep || true here temporarily to prevent exit during capture
+        matching_container=$(docker ps -a --format "{{.Status}}\t{{.Names}}"| grep "${CHECK_NAME}" || true) # Keep || true here, and below, to prevent exit during capture
         if [[ -z "${matching_container}" ]]; then
             echo "No containers found matching '${CHECK_NAME}'"
         else
@@ -107,7 +110,7 @@ check_docker_names() {
 		
         # Capture output, check if empty
         local running_containers
-        running_containers=$(docker ps --format "{{.Status}}\t{{.Names}}" | { grep -v NAMES || true; } | sort -n) # Isolate grep's potential failure
+        running_containers=$(docker ps --format "{{.Status}}\t{{.Names}}" | { grep -v NAMES || true; } | sort -n)
         if [[ -z "${running_containers}" ]]; then
             echo "No running containers found."
         else
@@ -117,7 +120,7 @@ check_docker_names() {
 		echo '====Other Docker Status Names====='
         # Capture output, check if empty
         local other_containers
-        other_containers=$(docker ps -a --format "{{.Status}}\t{{.Names}}" | grep -v Up || true) # Keep || true here temporarily to prevent exit during capture
+        other_containers=$(docker ps -a --format "{{.Status}}\t{{.Names}}" | grep -v Up || true)
         if [[ -z "${other_containers}" ]]; then
             echo "No containers found with other statuses."
         else
@@ -374,7 +377,7 @@ if [[ $# -eq 0 ]] || [[ "${FIRST_ARG}" == '--help' ]] || [[ "${FIRST_ARG}" == '-
   echo 'sweep-management:'
   echo "  cs, clean-sweep                    To run the entire dev tool suite"
   echo "  cst, clean-sweep-tests             To run only all the automated tests"
-  echo 'reset-mamagement:'
+  echo 'reset-management:'
   echo "  dr, dev-reset                      To reset OpenEMR only"
   echo "  di, dev-install                    To install OpenEMR (reset needs to be run prior)"
   echo "  dri, dev-reset-install             To reset and reinstall OpenEMR"
@@ -482,7 +485,8 @@ case "${FIRST_ARG}" in
     ;;
   *)
     # All other commands likely need a CONTAINER_ID
-    if [[ -z "${CONTAINER_ID}" ]] && [[ "${FIRST_ARG}" != '-d' ]]; then # Check if CONTAINER_ID is empty AND -d wasn't the *original* first arg
+    # Check if CONTAINER_ID is empty AND -d wasn't the *original* first arg
+    if [[ -z "${CONTAINER_ID}" ]] && [[ "${FIRST_ARG}" != '-d' ]]; then
       echo "Error: Could not automatically determine target OpenEMR container (tried '${INSANE_DEV_DOCKER}' and '${EASY_DEV_DOCKER}')." >&2
       echo "Use -d <container_name_or_id> to specify the target container." >&2
       exit 1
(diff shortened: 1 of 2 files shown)


msummers42:
@bradymiller Hello! I wrote this changeset hoping to start a conversation about some of the styles/conventions used in the bash scripts throughout the project. I would really appreciate your thoughts and any feedback on what is proposed here. Cheers!

utilities/openemr-cmd/openemr-cmd
```diff
@@ -1,16 +1,22 @@
 #!/bin/bash
+
+# Exit on error, unset var, pipe failure
+set -euo pipefail
```
kojiromike, line 4: Does anyone have thoughts about adding `-x` to this? I realize it can be noisy, but in my experience the only people who ever look at the output of a shell script actually want to know what's going on.

utilities/openemr-cmd/openemr-cmd
```diff
@@ -24,267 +30,292 @@ get_container_names() {
...
-FIRST_ARG=$1											                               # define the first parameter
-DEV_TOOLS=/root/devtools
+FIRST_ARG="${1:-}" # define the first parameter, default to empty if not set
+DEV_TOOLS='/root/devtools'
 
 check_docker_compose_install_or_not(){
-	DOCKER_COMPOSE_CODE=22
-	which docker-compose &>/dev/null
-	[ $? -ne 0 ] && echo "Please check docker-compose install or not." && exit $DOCKER_COMPOSE_CODE
+	local DOCKER_COMPOSE_CODE=22
```
kojiromike, line 56: I'll make a PR to bring the .editorconfig from openemr/openemr over so we can start to try to avoid mixing tabs and spaces, and other stuff.

utilities/openemr-cmd/openemr-cmd
```diff
@@ -24,267 +30,292 @@ get_container_names() {
...
-    fi
+  # Use the devtools function, passing all arguments ($@)
+  run_devtools_in_docker "${CONTAINER_ID}" set-swagger-to-multisite "$@"
 }
 
 set-webroot(){
-    if [ $# == 1 ]
-    then
+    local sed_command
+    if [[ $# == 1 ]]; then
```
kojiromike, line 289: Suggest using arithmetic context for a pure arithmetic check

```suggestion
    if (( $# == 1 )); then
```

utilities/openemr-cmd/openemr-cmd
```diff
@@ -24,267 +30,292 @@ get_container_names() {
...
-      if [ "$RESP" = "y" ]; then 
-        docker pull "$IMAGE";
-      else 
-        echo "Skipping $IMAGE";
+    local IMAGE RESP
+    for IMAGE in ${docker_images}; do
+      if [[ "${IMAGE}" != "IMAGE" ]]; then
+        # Use -r with read
+        read -r -e -p "Do you wish to pull ${IMAGE} ? (y) " RESP
+        if [[ "${RESP}" == "y" ]]; then
```
kojiromike, line 313:
```suggestion
        if [[ "${RESP,,}" = y ]]; then
```

the ,, makes it case insensitive, and the double `==` provides no advantage over, `=`. (IMO we should avoid bashisms unless they provide some advantage. Unfortunately once shellcheck knows this is bash, it doesn't flag `==`.)

utilities/openemr-cmd/openemr-cmd
```diff
@@ -294,15 +325,16 @@ FINAL_EXIT_CODE=0
...
-[ $? -ne 0 ] && echo "Please check docker install or not." && exit $DOCKER_CODE
+if ! command -v docker &>/dev/null; then
+  echo "Please check docker install or not."
+  exit "${DOCKER_CODE}"
+fi
 
 # Script usage.
-if [ $# -eq 0 ] || [ "$FIRST_ARG" == '--help' ] || [ "$FIRST_ARG" == '-h' ]
-then
+if [[ $# -eq 0 ]] || [[ "${FIRST_ARG}" == '--help' ]] || [[ "${FIRST_ARG}" == '-h' ]]; then
```
kojiromike, line 334:
```suggestion
if [[ $# -eq 0 || "${FIRST_ARG}" = '--help' || "${FIRST_ARG}" = '-h' ]]; then
```

utilities/openemr-cmd/openemr-cmd
```diff
@@ -321,125 +353,151 @@ then
...
-	echo "openemr-cmd $VERSION"
-	exit $VERSION_EXIT_CODE
+  echo "  ev, ensure-version                 Upgrade OpenEMR from specified old version to current version"
+  echo "  el, enable-ldap                    Turn on support for LDAP - login credentials are admin:admin"
+  echo "  dld, disable-ldap                  Turn off support for LDAP - standard login credentials"
+  echo "  ec, encoding-collation             Change the database character set and collation"
+
+  exit "${USAGE_EXIT_CODE}"
+
+elif [[ "${FIRST_ARG}" == '--version' ]] || [[ "${FIRST_ARG}" == '-v' ]]; then
```
kojiromike, line 436:
```suggestion
elif [[ "${FIRST_ARG}" = '--version' || "${FIRST_ARG}" = '-v' ]]; then
```

utilities/openemr-cmd/openemr-cmd
```diff
@@ -321,125 +353,151 @@ then
...
-elif [ "$FIRST_ARG" == '-d' ] && [ $# -lt 3 ]
-then
+elif [[ "${FIRST_ARG}" == '-d' ]] && [[ $# -lt 3 ]]; then
 	echo 'Please provide a docker id/name when using -d parameter.'
 	echo 'e.g. openemr-cmd -d docker_openemr-7-3-redis-312_1 dl'
-	exit $CHANGE_CONTAINER_CODE
+	exit "${CHANGE_CONTAINER_CODE}"
 else
   shift
 fi
```
kojiromike, line 468:
```suggestion
if [[ "${FIRST_ARG}" = '-d' ]]; then
  if (( $# < 3 )); then
    echo 'Please provide a docker id/name when using -d parameter.'
    echo 'e.g. openemr-cmd -d docker_openemr-7-3-redis-312_1 dl'
    exit "${CHANGE_CONTAINER_CODE}"
  fi
  CONTAINER_ID="$2"
  FIRST_ARG="$3"
  shift 3
else
  shift
fi
```

utilities/openemr-cmd/openemr-cmd
```diff
@@ -321,125 +353,151 @@ then
...
+# Check if a target container ID was determined or provided before proceeding with commands that need it
+# (Allow proceeding if the command doesn't require a container, like up, down, start, stop, --help, --version)
+case "${FIRST_ARG}" in
+  up|down|start|stop|--help|-h|--version|-v)
+    # These commands don't necessarily need a running container ID determined here
+    ;;
+  *)
+    # All other commands likely need a CONTAINER_ID
+    # Check if CONTAINER_ID is empty AND -d wasn't the *original* first arg
+    if [[ -z "${CONTAINER_ID}" ]] && [[ "${FIRST_ARG}" != '-d' ]]; then
```
kojiromike, line 489:
```suggestion
    if [[ -z "${CONTAINER_ID}" && "${FIRST_ARG}" != '-d' ]]; then
```

utilities/openemr-cmd/openemr-cmd
```diff
@@ -24,267 +30,292 @@ get_container_names() {
...
 copy_capsule_from_docker_to_host(){
-	BACKUP_FILE=$1
-	BACKUP_HOST_DIR=$2 #optional parameter
-	BACKUP_FILE_CODE=19
-	if [ $# != 1 ] && [ $# != 2 ]
-	then
+	local BACKUP_FILE="$1"
+	local BACKUP_HOST_DIR="${2:-}" #optional parameter, default to empty
+	local BACKUP_FILE_CODE=19
+	if [[ $# != 1 ]] && [[ $# != 2 ]]; then
```
kojiromike, line 172:
```suggestion
	if (( $# != 1 && $# != 2 )); then
```

(Although if I had my druthers we'd flip negative if expressions and this'd start with `(( $# == 1 || $# == 2 ))`.)

utilities/openemr-cmd/openemr-cmd-h
```diff
@@ -69,5 +76,9 @@ case "$FIRST_ARG" in
 		openemr-cmd -h
 		;;
 		*)
-		openemr-cmd -h|grep -i $FIRST_ARG
+		# Quoted variable
+		openemr-cmd -h | grep -i "${FIRST_ARG}"
+		;;
 esac
+
+exit 0 # Explicit success exit at end of script
```
kojiromike, line 84: Was this a shellcheck suggestion? It seems unnecessary, as the script will always exit early if there's an error (due to `set -e` at the top).

utilities/openemr-cmd/openemr-cmd-h
```diff
@@ -1,13 +1,18 @@
...
-	openemr-cmd -h|grep $FILTER_ARG -A $LINE_COUNT
+	local FILTER_ARG="$1"
+	local LINE_COUNT="$2"
+	# Quoted variables
+	openemr-cmd -h | grep "${FILTER_ARG}" -A "${LINE_COUNT}"
 }
 
-if [ $# -eq 0 ]
+# Use [[ ]] for test
+if [[ $# -eq 0 ]]
```
kojiromike, line 15:
```suggestion
# Use (( )) for numeric comparisons
if (( $# == 0 ))
```

utilities/openemr-cmd/openemr-cmd
```diff
@@ -496,190 +554,191 @@ case "$FIRST_ARG" in
...
 		;;
-
 	*)
-		$DOCKER_EXEC_CMD $CONTAINER_ID $SHELL_TYPE "$DEV_TOOLS $FIRST_ARG"
+		# Default case: assume it's a devtools command
+		run_devtools_in_docker "${CONTAINER_ID}" "${FIRST_ARG}" "$@"
 		;;
 esac
-exit $FINAL_EXIT_CODE
+exit "${FINAL_EXIT_CODE}"
```
kojiromike, line 744: With `set -e` it's hard to be sure the script will exit with exactly this code unless we do a lot more review.

utilities/openemr-cmd/openemr-cmd
```diff
@@ -321,125 +353,151 @@ then
...
+      exit 1
+    fi
+    # If -d was used, CONTAINER_ID was set in the parameter parsing block
+    ;;
+esac
 
 # See how we were called.
 # For the shift usage, it used to cover the insane env.
-case "$FIRST_ARG" in
+case "${FIRST_ARG}" in
```
kojiromike, line 500: This `case` expression looks like it would benefit from an associative array.

utilities/openemr-cmd/openemr-cmd
msummers42: Just made a change that should address that specific area. Lots of mixing of tabs & spaces in here.

utilities/openemr-cmd/openemr-cmd
msummers42: I just made a style change, not logic.

utilities/openemr-cmd/openemr-cmd
msummers42: I agree, but `set -e` was in place already and I simply added the unbound var and pipefail bits. I think it's probably OK for the moment.

utilities/openemr-cmd/openemr-cmd-h
msummers42: I just added that to be clear what exist status was used since there are a bunch of custom statuses set in various places in the scripts. It's not necessary.

utilities/openemr-cmd/openemr-cmd
```diff
@@ -24,267 +30,292 @@ get_container_names() {
...
-      read -e -p "Do you wish to pull $IMAGE ? (y) " RESP
-      if [ "$RESP" = "y" ]; then 
-        docker pull "$IMAGE";
-      else 
-        echo "Skipping $IMAGE";
+    local IMAGE RESP
+    for IMAGE in ${docker_images}; do
+      if [[ "${IMAGE}" != "IMAGE" ]]; then
+        # Use -r with read
+        read -r -e -p "Do you wish to pull ${IMAGE} ? (y) " RESP
```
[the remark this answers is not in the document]
msummers42: Just using what was there, though I added the `-r` to avoid escape garbage.

utilities/openemr-cmd/openemr-cmd
msummers42: It's very noisy, and I suggested adding that when debugging versus standard usage. It might be interesting to add a debug flag that would `set -x`. I was thinking about adding one of the standard `die_msg()` that would print more debugging info at some point.

utilities/openemr-cmd/openemr-cmd
msummers42: Can you retry this suggestion, it is giving me an error "Applying suggestions on deleted lines is currently not supported. "

utilities/openemr-cmd/openemr-cmd
```diff
@@ -294,15 +347,16 @@ FINAL_EXIT_CODE=0
...
 
 # Script usage.
-if [ $# -eq 0 ] || [ "$FIRST_ARG" == '--help' ] || [ "$FIRST_ARG" == '-h' ]
-then
+if [[ $# -eq 0 || "${FIRST_ARG}" = '--help' || "${FIRST_ARG}" = '-h' ]]; then
   echo ""
-  echo "Usage: `basename $0` COMMAND [ARGS]"
-  echo "Usage: `basename $0` -d <docker name> COMMAND [ARGS]"
+  echo "Usage: $(basename "$0") COMMAND [ARGS]"
+  echo "Usage: $(basename "$0") -d <docker name> COMMAND [ARGS]"
```
kojiromike, line 359:
```suggestion
  echo "Usage: ${0##*/} COMMAND [ARGS]"
  echo "Usage: ${0##*/} -d <docker name> COMMAND [ARGS]"
```

…to avoid the subshell and avoid doing it twice :)

msummers42: I tend to forget about this way to grab the basename, nice.

jesdynf:
I'll look this over no later than Saturday.

jesdynf:
@bradymiller I've sight-read this and it all looks good, but I haven't tested it and I don't do anything that *would* test it, nor am I certain I understand who does use these, and where they do so, so I'm not this-minute willing to pull it in. Do you think it's probably fine and a problem would just break developer environments until they yell loud enough for us to fix it?

bradymiller:
hi @jesdynf @msummers42 , I'll do road testing on this on Tuesday. If goes well (which I am guessing it well), will then bring this in.

bradymiller:
Sanity testing well on both easy and insane dev environments. Bringing in. Thanks for the improvements @msummers42 ! and thanks for the reviews @kojiromike and @jesdynf !
Especially like the new support for `docker compose` and continued compatibility with `docker-compose` !

PR #417, opened 2025-04-25
shellcheck and fixup unlock_admin.sh scripts

kojiromike:

Fix the issues identified by shellcheck for this script (same script in various places).

- SC2164: Check if `cd` was successful before continuing.
- SC2086: Double quote to prevent globbing and word splitting.

### DIFF
--- a/docker/openemr/5.0.0/utilities/unlock_admin.sh
+++ b/docker/openemr/5.0.0/utilities/unlock_admin.sh
@@ -1,4 +1,3 @@
 #!/bin/sh
 
-cd /root
-php ./unlock_admin.php $1
+cd /root && php ./unlock_admin.php "$1"
--- a/docker/openemr/5.0.1/utilities/unlock_admin.sh
+++ b/docker/openemr/5.0.1/utilities/unlock_admin.sh
@@ -1,4 +1,3 @@
 #!/bin/sh
 
-cd /root
-php ./unlock_admin.php $1
+cd /root && php ./unlock_admin.php "$1"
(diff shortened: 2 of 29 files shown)


jesdynf:
So there's two things I wanted to ask about here.

One, it doesn't look like shellcheck is spotting any of the sibling scripts in the `docker` subdirectories -- it's spotted the versioned flex-* scripts, but it's missed the bare "7.0.3"-style directories. Was this intended? Should it be otherwise?

Two, I agree with the quoting, but "check for directory change success" seems really weird to me. It works here, but it wouldn't work for multii-line scripts -- or even a script with a single line beyond this. Is shellcheck going to want us to wrap every cd in some sort of test, or is this some kind of special case that only shows up for the shortest of scripts?

kojiromike:
> It doesn't look like shellcheck is spotting any of the sibling scripts in the `docker` subdirectories -- it's spotted the versioned flex-* scripts, but it's missed the bare "7.0.3"-style directories. Was this intended? Should it be otherwise?

Oh, that was my miss, not shellcheck's. I've fixed it. Note that I've programmed the action only to check edited scripts in a PR.

> Two, I agree with the quoting, but "check for directory change success" seems really weird to me. It works here, but it wouldn't work for multii-line scripts -- or even a script with a single line beyond this. Is shellcheck going to want us to wrap every cd in some sort of test, or is this some kind of special case that only shows up for the shortest of scripts?

You're right that `cd x && …` is only well suited to short handoffs like these, but we definitely do want an equivalent check for longer scripts. That said, this check is also satisfied with `set -e/errexit` mode.

PR #421, opened 2025-04-30
chore: address shellcheck findings in `xrecovery-final.sh`

kojiromike:

Address the issues uncovered by shellcheck in one set of (identical) files.

- Add quotes around variables to avoid accidental wordsplitting and glob expansion.
- Add curly braces around variables for consistency and readability.
- Combine `touch` to create two files.
- Remove `exit 0` to exit with the correct exit code
- Use `set -euo pipefail` to catch errors earlier.

### DIFF
--- a/docker/mysql-xtrabackup-final/xrecovery-final.sh
+++ b/docker/mysql-xtrabackup-final/xrecovery-final.sh
@@ -1,24 +1,24 @@
 #!/bin/bash
 
+set -euo pipefail
+
 WORKDIR=/mnt/backups/recovery
 
 if service mysql status; then
-  echo xrecovery-final: mysql must not be running, yet it is; aborting
+  echo 'xrecovery-final: mysql must not be running, yet it is; aborting'
   exit 1
 fi
 
 # Even though MySQL isn't running, I can't just make off with /var/lib/mysql,
 # since the container mounts it.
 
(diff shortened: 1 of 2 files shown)


jesdynf:
I'm willing to approve this, but can you walk me through how to turn off the carping for unprotected *echo* statements? That's just being fiddly.

kojiromike:
> can you walk me through how to turn off the carping for unprotected _echo_ statements? That's just being fiddly.

shellcheck doesn't actually seem to care if you quote around echo statements if a variable is not involved. I did those myself, out of habit. I stand by it, though, because echo will have less surprising behavior if its argument is always quoted.

jesdynf:
I'd prefer not to accept the edits to the echo statements unless we come to a decision that that's a style change we're actually going to enforce going forward, and this minute I'm not planning to do that.

kojiromike:
Actually the script doesn't work correctly without the quotes. Here's one example from the script:

```
$ echo xrecovery-final: mysql must not be running, yet it is; aborting
xrecovery-final: mysql must not be running, yet it is
bash: aborting: command not found
```

It might be barely noticeable since the very next line is `exit 1`, but it is, in any case, clearly not what the code is intended to do.

jesdynf:
I will let this one through to pick up the work done.

PR #425, opened 2025-05-18
more granular docker tags

bradymiller:
Adding a tag with patch and also adding a tag with patch-date to the production docker releases.
This is so that orchestrations that need to ensure never change images on the fly (ie. they can pin to the patch-date tag)


bradymiller:
expediting this since need for the patch release today, so bringing in.

PR #431, opened 2025-06-10
ci: test that builds work

### LINKED ISSUE #430: Test Docker Images
In addition to the testing that already goes on in openemr/openemr, I think we should test that Docker images built from this repo are essentially healthy. Mainly, if we start making changes to the build process, openemr.sh, etc, this will provide an indication of success before we encounter issues in the other repo.

kojiromike:
Fixes #430

Add testing action around the prod builds, so we know if we break them.

Add a Github Action workflow for testing 6.1.0 and 7.0.0 and up.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -1,0 +1,39 @@
+# OpenEMR Docker Testing
+
+This directory contains a GitHub Actions workflow to test the OpenEMR 7.0.4 Docker image.
+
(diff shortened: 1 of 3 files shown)


jesdynf:
This looks good to me; the only reason I don't pull it in is that I don't know how GitHub Action billing or quotas work -- @bradymiller , is this unobjectionable?

PR #434, opened 2025-06-10
chore(docker/openemr): obsolete old flex versions

### LINKED ISSUE #433: Move obsolete flex images
In #422, we moved obsoleted versioned/prod images so it's easy to tell what's maintained and what's historical. We need to do the same for the flex images.

kojiromike:
Fixes #433

Makes it clear that some old versions of OpenEMR are not maintained or expected to build anymore.

Move all the flex images that are not used in https://github.com/openemr/openemr/blob/ba01bd31e91c61afa745c8d14271b074b58d0e3e/docker/development-insane/docker-compose.yml into docker/openemr/obsolete.


bradymiller:
I'm ok with getting more aggressive on this and ok with this PR. This will at least test the flexes that are documented to use for easy dev (to test php versions 8.2, 8.3, and 8.4). And can still build the more recent obsoletes if really needed.
Would also get more aggressive on the production dockers and drop at least 6.1.0.

PR #435, opened 2025-06-10
ci: test flex images

### LINKED ISSUE #432: Test Flex Images
In #430, we added testing for the versioned/prod images. We should add tests for the flex images as well. This is a bit different than #430 because the flex images don't ship the code.

kojiromike:
Fixes #432

Ensure the flex image builds are tested.

- Add github actions for testing the flex images.


msummers42:
Can support for the EOL'd versions of Alpine be dropped? 3.18 is EOL and unsupported except by request as of May 9th 2025, while 3.17 is EOL and unsupported except by request as of Nov 22 2024.

https://alpinelinux.org/releases/

kojiromike:
> Can support for the EOL'd versions of Alpine be dropped? 3.18 is EOL and unsupported except by request as of May 9th 2025, while 3.17 is EOL and unsupported except by request as of Nov 22 2024.
> 
> https://alpinelinux.org/releases/

I hope we can drop them. The thing that's failing for the two old versions is in platform_check.php. In 3.15, it's failing because php is 8.0.1 and composer only supports 8.1.0 and up. In 3.17 the flex container appears to be running a 32-bit version of php and composer only supports 64-bit php now. I'm going to remove support for those from my tests and recommend we move them to obsolete.

bradymiller:
hi @kojiromike , brought in the PR to obsolete 15 and 17. Any more changes needed on this PR (noted the hard-coding of versions in matrix's which now guessing can gather from directory) before bringing in?

kojiromike:
(Just to say there are versions of flex-* in docker/openemr/obsolete now that do pass tests against some older versions of openemr, and do _not_ pass tests against 703 and 704.)

bradymiller:
Probably would just test all flex's against master since the production
docker build tests will take care of the pertinent rel-70* branches.

kojiromike:
> Probably would just test all flex's against master since the production docker build tests will take care of the pertinent rel-70* branches.

OK, done

PR #437, opened 2025-06-13
fix(flex-edge): shellcheck complaints

### LINKED ISSUE #436: Fix Shellcheck findings in flex-edge builder
There are lots of shellcheck findings for openemr.sh and devtoolsLibrary.source. I will attempt to address the ones for flex-edge first.

kojiromike:
Fixes #436

Address the shellcheck findings for openemr.sh and utilities/devtoolsLibrary.source in docker/openemr/flex-edge

- Quote parameter expansions 
- Ensure default values are set
- Replace backticks with `$()`
- Replace expr with `$(( ))` arithmetic syntax
- Use `/bin/sh` instead of `/bin/bash` so both scripts are consistent


ashermcs:
On eyeballing it nothing seems like a problem, but I don't understand why backticks are bad now. What's going on there?

jesdynf:
On eyeballing it I don't find the changes problematic, but I don't understand why backticks are bad now. What's going on there?

kojiromike:
@jesdynf Backticks still work, but they are a really antiquated way to do command substitution, and have several problems. https://mywiki.wooledge.org/BashFAQ/082 explains it more completely than I can.

jesdynf:
This is what I get for thirty years of shell scripting, I suppose. Well enough, thank you.

jesdynf:
Looking nice, thanks!

docker/openemr/flex-edge/openemr.sh
```diff
@@ -85,27 +97,26 @@ auto_setup() {
...
     set -o noclobber
-    { > /var/www/localhost/htdocs/openemr/sites/docker-leader ; } &> /dev/null || AUTHORITY=no
+    { : > /var/www/localhost/htdocs/openemr/sites/docker-leader ; } > /dev/null 2>&1 || AUTHORITY=no
     set +o noclobber
-    
-    if [ "$AUTHORITY" == "no" ] &&
-       [ ! -f /var/www/localhost/htdocs/openemr/sites/docker-completed ]; then
-        while swarm_wait; do
+
+    if [ "${AUTHORITY}" = "no" ]; then
```
bradymiller, line 112: shouldn't the quotes be removed from the no above (noted this pattern of change elsewhere)?

bradymiller: seeing lots of other places below where quotes is not removed. maybe i am missing something?

bradymiller:
Also note very minimal difference between this and flex-3.21, so can basically copy over files from flex-edge to there (and guessing the other flexes albeit those others will likely be obsolete in relatively near future).
For example, here is diff from flex-3.21 to flex-edge:
```
[03:19][brady2@pop-os:~/git/openemr-devops/docker/openemr(master)]$ diff -r flex-3.21 flex-edge
diff -r flex-3.21/Dockerfile flex-edge/Dockerfile
1c1
< FROM alpine:3.21
---
> FROM alpine:edge
Only in flex-edge: .dockerignore
Only in flex-edge: .gitignore
diff -r flex-3.21/php.ini flex-edge/php.ini
919d918
< 
diff -r flex-3.21/README.md flex-edge/README.md
8a9
>   - If you are using Raspberry Pi, then change the `mariadb:10.5` to `jsurf/rpi-mariadb`.
27c28
<     image: openemr/openemr:flex-3.21
---
>     image: openemr/openemr:flex-edge
50c51
< [![Try it!](https://github.com/play-with-docker/stacks/raw/cff22438cb4195ace27f9b15784bbb497047afa7/assets/images/button.png)](http://play-with-docker.com/?stack=https://gist.githubusercontent.com/bradymiller/320b83993912a72b73212f56a17e668c/raw/5bf5c78e61e69ca713b0f716d4397536fde14fbf/openemr-flex-321-docker-example-docker-compose.yml)
---
> [![Try it!](https://github.com/play-with-docker/stacks/raw/cff22438cb4195ace27f9b15784bbb497047afa7/assets/images/button.png)](http://play-with-docker.com/?stack=https://gist.githubusercontent.com/bradymiller/5b654d0c6fd996332dc604156680e1da/raw/72255a577d49e56076bf1cb0e089796550b8a7d0/openemr-flex-edge-docker-example-docker-compose.yml)
58c59
< 64 bit architecture is supported on Raspberry Pi.
---
> Both 32 bit and 64 bit architectures are supported on Raspberry Pi. If you are using Raspberry Pi, then need to change the `mariadb:10.5` to `jsurf/rpi-mariadb` in the above docker-compose.yml example.

```

bradymiller:
hi @kojiromike , This is breaking in easy dev flex environment (just keep getting the loop message in log that unable to connect to mysql container). Noted this when working on #440 (when i reverted these changes, it then worked; obviously not a long term work around in that PR, but at least will keep things moving on that PR until things are sorted out).

kojiromike:
@bradymiller aha, thanks, I see what happened and opened #441.

docker/openemr/flex-edge/openemr.sh
kojiromike: If this were bash, then we would use `[[ ... ]]` and remove the quotes. But here in POSIX sh we want the quotes, because they prevent improper expansions that will make the test fail.

kojiromike:
@bradymiller @jesdynf that said, can you help me understand how the tests passed in this case? I gather there must be another bug here besides #441, one that makes the test setup here different from the ones in openemr/openemr.

docker/openemr/flex-edge/openemr.sh
bradymiller: hi, but then why were the quotes removed above in lines 81 and 106?

kojiromike: Ah, I removed the quotes around bare words with no special characters because I think they're less ambiguous that way. It wouldn't hurt to leave them in single quotes, although I do think double quotes should be reserved for when they're actually necessary.

bradymiller: sounds good

PR #442, opened 2025-06-16
fix(flex-edge/auto_configure.php): wordsplit -f argument

### LINKED ISSUE #441: CONFIGURATION is not properly supplied to mysql command in flex-edge
#437 introduced a bug by quoting the expansion of `"${CONFIGURATION}"` in openemr.sh and devtoolsLibrary.source in flex-edge. The previous version of the script used an unquoted expansion, relying on the shell to wordsplit the mysql connection string before passing it to `auto_configure.php`. This is problematic because unquoted expansions risk the shell doing undesirable interpretations of the string.

A better solution would be to allow `auto_configure.php` to interpret CONFIGURATION directly.

kojiromike:
Fixes #441

Make `auto_configure.php` handle splitting the mysql connection string,
since we don't want to rely on the shell to do it.

Make `auto_configure.php` handle splitting the mysql connection string,
since we don't want to rely on the shell to do it.

### DIFF
--- a/docker/openemr/flex-edge/auto_configure.php
+++ b/docker/openemr/flex-edge/auto_configure.php
@@ -26,10 +26,25 @@ $installSettings['no_root_db_access']        = 'BLANK';
 $installSettings['development_translations'] = 'BLANK';
 // Collect parameters(if exist) for installation configuration settings
 for ($i=1; $i < count($argv); $i++) {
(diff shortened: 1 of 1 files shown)


bradymiller:
hi @kojiromike ,
Gonna bring this in and the other stuff fixing up flex-edge to keep things going (and can then incorporate the changes in the generic flex docker stuff am working on).

PR #452, opened 2025-06-25
fix(shell): address SC2250

kojiromike:
Ran `shellcheck -i SC2250 -f diff | git apply`


jesdynf:
Seems reasonable, and the shellcheck errors aren't coming from it.

PR #453, opened 2025-06-25
fix(shell): shellcheck SC3014

kojiromike:


kojiromike:
> @kojiromike, approving the other shellcheck patch forced this PR into conflicts. Can you resolve it, or do I need to try?

I'll take care of it in a minute here

PR #459, opened 2025-07-01
fix(shell): quote variables for SC2086/SC2248/SC2046

kojiromike:

Fix shellcheck SC2086, SC2248, and SC2046 (variable-quoting) findings across the shell files checked in CI. This is the first of several focused PRs chipping away at the full shellcheck backlog — remaining codes (SC2044 find-loops, SC2154 unset vars, SC2310/SC2312 set -e behavior, etc.) are queued for follow-up PRs.

After this PR, `shellcheck --include=SC2086,SC2248,SC2046` is clean on every file the CI workflow scans.

- Quote variables in `docker/openemr/{7.0.4,8.0.0}/{openemr,ssl}.sh`, the `fsupgrade-*.sh` scripts, `packages/standard/{ami,scripts}/*.sh`, and the three `utilities/` installer/migrator/monitor scripts.
- Where word-splitting is intentional, add targeted per-line `# shellcheck disable=SC2086` with a reason comment:
  - `openemr.sh`: `CONFIGURATION` intentionally splits into multiple `-f` flags (see #539).
  - `ssl.sh`: `EMAIL` intentionally splits to pass `-m address` as separate args.
- In `fsupgrade-*.sh` scripts, remove redundant `if [ ! -d X ]; then mkdir -p X; fi` guards (`mkdir -p` is already idempotent) and collapse the 13 sequential `mkdir -p` calls into a single multi-arg `mkdir -p`.
- Re-sync `docker/openemr/8.0.0/upgrade/fsupgrade-{1..7}.sh` from the updated `7.0.4/` copies so the CI content-hash dedup still collapses byte-identical files.

CI's shellcheck workflow remains failing — those failures are all from SC codes queued for follow-up PRs, not from any fix in this change.

#### AI disclosure:

### DIFF
--- a/docker/openemr/7.0.4/openemr.sh
+++ b/docker/openemr/7.0.4/openemr.sh
@@ -29,7 +29,7 @@ auto_setup() {
 
     #create temporary file cache directory for auto_configure.php to use
     TMP_FILE_CACHE_LOCATION="/tmp/php-file-cache"
-    mkdir ${TMP_FILE_CACHE_LOCATION}
+    mkdir "${TMP_FILE_CACHE_LOCATION}"
 
     #create auto_configure.ini to be able to leverage opcache for operations
     touch auto_configure.ini
@@ -42,10 +42,11 @@ auto_setup() {
     echo "opcache.max_accelerated_files=1000000" >> auto_configure.ini
 
     #run auto_configure
+    # shellcheck disable=SC2086 # CONFIGURATION intentionally word-splits into multiple -f flags (see openemr/openemr-devops#539)
     php auto_configure.php -c auto_configure.ini -f ${CONFIGURATION} || return 1
 
     #remove temporary file cache directory and auto_configure.ini
-    rm -r ${TMP_FILE_CACHE_LOCATION}
+    rm -r "${TMP_FILE_CACHE_LOCATION}"
     rm auto_configure.ini
 
     echo "OpenEMR configured."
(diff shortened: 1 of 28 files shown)


docker/openemr/7.0.3/openemr.sh
```diff
@@ -42,7 +42,7 @@ auto_setup() {
     echo "opcache.max_accelerated_files=1000000" >> auto_configure.ini
 
     #run auto_configure
-    php auto_configure.php -c auto_configure.ini -f ${CONFIGURATION} || return 1
+    php auto_configure.php -c auto_configure.ini -f "${CONFIGURATION}" || return 1
```
bradymiller: Will this require this change (both for 7.0.3 and 7.0.4) if not already there:
https://github.com/openemr/openemr-devops/blob/8f3eca7df00d900298e7346cba7cbb086ae47c1b/docker/openemr/flex-edge/auto_configure.php#L27-L48

kojiromike:
bah, there was a bug in my git ls-files and this didn't actually catch all the shellcheck issues. There are still a ton in the xbackup scripts.

jesdynf:
Those might not be worth fighting. I need to migrate away from the current MySQL image with Percona's XtraBackup stuff to something with MariaDB.

jesdynf:
@kojiromike I've migrated off the xbackup stuff, which was blocking you, but this pull is aging -- should we close or can we move on it?

jesdynf:
Looks okay, but a lot of the constructs aren't ones I use regularly. Which, I suppose, is the point of the commit.

PR #464, opened 2025-07-04
Add in the auto_configure.php to root

### LINKED ISSUE #463: feat: Put auto_configure.php in /root/ folder so non-easy dev setup dockers can do snapshot & restore
### Describe the problem

 The devtools for restore / backup require the auto_configure.php to be
    in the /root/ folder.  The auto_configure.php only gets copied to root
    via the openemr.sh file if its an easy dev docker setup.  However this
    messes up the ci installations (such as inferno) which uses the
    restore/backup in a non-easy dev type setup (where the codebase is not
    shared).
   
Just copying auto_configure.php to root solves that issue.

adunsulag:
The devtools for restore / backup require the auto_configure.php to be in the /root/ folder.  The auto_configure.php only gets copied to root via the openemr.sh file if its an easy dev docker setup.  However this messes up the ci installations (such as inferno) which uses the restore/backup in a non-easy dev type setup (where the codebase is not shared).

Just copying auto_configure.php to root solves that issue.

Fixes #463

### DIFF
--- a/docker/openemr/flex/Dockerfile
+++ b/docker/openemr/flex/Dockerfile
@@ -122,6 +122,8 @@ COPY utilities/devtoolsLibrary.source /root/
 RUN mkdir /snapshots
 RUN mkdir /certs
 RUN mkdir -p /couchdb/original
+#Copy auto_configure so we can do backup/restore with dev tools
+COPY auto_configure.php /root/
 #Copy demo data to root
 COPY utilities/demo_5_0_0_5.sql /root/
 RUN chmod 500 /root/devtools


docker/openemr/flex/Dockerfile
```diff
@@ -122,6 +122,8 @@ COPY utilities/devtoolsLibrary.source /root/
 RUN mkdir /snapshots
 RUN mkdir /certs
 RUN mkdir -p /couchdb/original
+#Copy auto_configure so we can do backup/restore with dev tools
```
bradymiller, line 125: also clarify in comment that we are aware are also adding above (line 113) to another place (just so we don't remove one of them in future)

bradymiller: rec removing lines 432-433 in openemr.sh to clean up loose end

PR #467, opened 2025-07-09
fix(shell): all shellcheck issues (except quotes)

kojiromike:
Resolved the following shellcheck warnings and errors:

- SC2006: Use $(...) instead of legacy `...` backticks
- SC2016: Expressions don't expand in single quotes (disabled where intentional)
- SC2029: Note that client-side expansion happens (disabled where intentional)
- SC2086: Double quote to prevent globbing and word splitting
- SC2162: read without -r will mangle backslashes
- SC2001: See if you can use ${variable//search/replace} instead of sed
- SC2004: $/${} is unnecessary on arithmetic variables
- SC2143: Use grep -q instead of comparing output
- SC2009: Consider using pgrep instead of ps|grep
- SC2181: Check exit code directly with e.g. 'if mycmd;', not '$?'
- SC2034: Variable appears unused (verify it or export if used externally)
- SC2119: Use function_name "$@" if function's $1 should mean script's $1
- SC2120: Function references arguments, but none are ever passed
- SC2154: Variable is referenced but not assigned
- SC2015: Note that A && B || C is not if-then-else
- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined
- SC2003: expr is antiquated. Consider rewriting this with $((..))
- SC2103: Use a ( subshell ) to avoid having to cd back

Fixed issues in:
- utilities/openemr-env-migrator/openemr-env-migrator
- utilities/openemr-env-installer/openemr-env-installer
- utilities/openemr-monitor/monitor-installer
- docker/openemr/7.0.3/utilities/devtoolsLibrary.source
- docker/openemr/7.0.4/utilities/devtoolsLibrary.source

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/docker/openemr/7.0.3/utilities/devtoolsLibrary.source
+++ b/docker/openemr/7.0.3/utilities/devtoolsLibrary.source
@@ -1,6 +1,18 @@
 #!/bin/sh
 
 prepareVariables() {
+    # Set default values for environment variables if not provided
+    : "${MYSQL_DATABASE:=''}" \
+      "${MYSQL_HOST:=localhost}" \
+      "${MYSQL_PASS:=''}" \
+      "${MYSQL_PORT:=''}" \
+      "${MYSQL_ROOT_PASS:=''}" \
+      "${MYSQL_ROOT_USER:=''}" \
+      "${MYSQL_USER:=''}" \
+      "${OE_PASS:=''}" \
+      "${OE_USER:=''}" \
+      "${SQL_DATA_DRIVE:=''}"
+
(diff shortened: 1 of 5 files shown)


ashermcs:
Yeah, sorry, it's probably fine. I wanted to look at this closer but a lot of life happened this weekend.

jesdynf:
Yeah, sorry, it's probably fine. I wanted to look at this closer but a lot of life happened this weekend

bradymiller:
Sounds good. gonna bring in to keep the wheels moving :)

PR #483, opened 2025-09-24
pick up new docker-compose standards

jesdynf:

new version of Alpine or Docker changes container naming schema

### DIFF
--- a/packages/express/ami/ami-config.sh
+++ b/packages/express/ami/ami-config.sh
@@ -13,7 +13,7 @@ f () {
         sleep 5
     done
 
-    until docker top $(docker ps | grep _openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
+    until docker top $(docker ps | grep -- -openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
     do
         echo "waiting for service start..."
         sleep 20
--- a/packages/express/ami/ami-rekey.sh
+++ b/packages/express/ami/ami-rekey.sh
@@ -38,16 +38,16 @@ do
     sleep 5
 done
 
-until docker top $(docker ps | grep _openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
+until docker top $(docker ps | grep -- -openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
 do
     sleep 3
 done
 
 # reset password
-docker exec $(docker ps | grep _openemr | cut -f 1 -d " ") /root/unlock_admin.sh $(curl http://169.254.169.254/latest/meta-data/instance-id)
+docker exec $(docker ps | grep -- -openemr | cut -f 1 -d " ") /root/unlock_admin.sh $(curl http://169.254.169.254/latest/meta-data/instance-id)
 
 # reset SSL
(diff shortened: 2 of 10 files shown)


packages/express/ami/ami-config.sh
```diff
@@ -13,7 +13,7 @@ f () {
         sleep 5
     done
 
-    until docker top $(docker ps | grep _openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
+    until docker top $(docker ps | grep -- -openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
```
kojiromike:
```suggestion
    until docker top $(docker ps | awk '/-openemr/{ print $1 }') | grep -qF httpd
```

Although if `jq` is available, `docker ps --format json` could be even cleaner.

packages/express/ami/ami-rekey.sh
```diff
@@ -38,16 +38,16 @@ do
...
 
-until docker top $(docker ps | grep _openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
+until docker top $(docker ps | grep -- -openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
 do
     sleep 3
 done
 
 # reset password
-docker exec $(docker ps | grep _openemr | cut -f 1 -d " ") /root/unlock_admin.sh $(curl http://169.254.169.254/latest/meta-data/instance-id)
+docker exec $(docker ps | grep -- -openemr | cut -f 1 -d " ") /root/unlock_admin.sh $(curl http://169.254.169.254/latest/meta-data/instance-id)
```
kojiromike: Just checking, are these all supposed to be running in docker compose? If so, instead of using `docker ps | grep` could we use `docker compose exec …`?

jesdynf:
From the host machine, we're hunting for the container docker compose made,
then we're running the script we named in that container with the exec.
*Is* there a better way to do that?

Asher Densmore-Lynn ***@***.***>

packages/express/ami/ami-config.sh
```diff
@@ -13,7 +13,7 @@ f () {
         sleep 5
     done
 
-    until docker top $(docker ps | grep -- -openemr | cut -f 1 -d " ") | grep httpd &> /dev/null
+    until docker top $(docker ps | awk '/-openemr/{ print $1 }') | grep -qF httpd
```
kojiromike: If this is actually docker compose, we can do…
```suggestion
    until docker compose top openemr | grep -qF httpd
```

kojiromike: But also, we added healthchecks to the services, so `docker compose up -d --wait` will block until the services are up…

jesdynf: WHEN DID DOCKER COMPOSE LEARN TO--

Okay, I'm not mad.

jesdynf: Wait, where is this healthcheck? I see one in the compose in 7.0.4, is that what you're taking about?

kojiromike: https://github.com/openemr/openemr-devops/blob/57415ac7175bd2f0eea924a6e59da8c199d0f393/docker/openemr/compose.yml#L9

jesdynf: Okay, so that's not part of the container itself, we'll have to port that awareness to any composes that use it. Which can be done, I'll just have to weave that through all the everythings in /packages.

PR #485, opened 2025-09-24
chore: pin 7.0.3 docker build to specific tag v7_0_3_4

bradymiller:
chore: pin 7.0.3 docker build to specific tag v7_0_3_4


bradymiller:
Bringing this in. then will build 7.0.3 docker.

PR #498, opened 2025-11-16
fix(openemr.sh): invalid sed to inject session path

kojiromike:
Fixes openemr/openemr#9473

Replaces the invalid sed by creating a php.ini include instead of using sed.


jesdynf:
Looks good to me. I'm not sure how long it's been since I tried building anything novel and pulled in Redis.

bradymiller:
neat stuff. Would also fix in 7.0.4 docker.

bradymiller:
i just rebuilt the flex-3.22 docker set

PR #499, opened 2025-11-18
MariaDB backup manager

jesdynf:
adds agent for mariadb-backup for consistent database backups

### DIFF
--- a/utilities/mariadb-backup-manager/restore-client/restore.sh
+++ b/utilities/mariadb-backup-manager/restore-client/restore.sh
@@ -26,7 +26,7 @@ EOF
 
 pickLatestBackup() {
     TIMESTAMP=$(find . -maxdepth 1 -type f -name '*.manifest' -printf '%f\n' | \
-        sort | head -n 1 | sed -E 's/(.*)\.manifest/\1/')
+        sort -r | head -n 1 | sed -E 's/(.*)\.manifest/\1/')
     if [[ $? -ne 0 || -z "${TIMESTAMP}" ]]; then
         echo "failure, cannot autodetect manifest"
     fi


bradymiller:
2DV

utilities/mariadb-backup-manager/README.md
jesdynf: @bradymiller @stephenwaite I don't need a full review, but could I trouble you to glance at the readme? Does it seem like it covers what the user might want to know?

jesdynf:
We're all green, thankfully. Final testing tomorrow to make sure it still *works* after I dry-patched a bunch of expressions.

utilities/mariadb-backup-manager/backup-client/backup.sh
```diff
@@ -9,7 +9,8 @@ TIMESTAMP="$(date +%Y-%m-%d-%H-%M-%S)"
 
 BACKUPCLASS=undecided
 considerState() {
-    LASTFULLBACKUP="$(cd "${BACKUPVOLUME_TARGET}"; ls -r *.manifest | head -n 1 | sed -E "s/(.*)\.manifest/\1/")"
+    LASTFULLBACKUP="$(find "${BACKUPVOLUME_TARGET}" -type f -name '*.manifest' -printf '%f\n' | \
```
kojiromike: You could use `shopt -s globstar` here instead of find and create an array of all the .manifest files. Then potentially reuse that down in `TARGET_MANIFEST` as well. Doing it that way you can probably use parameter expansion instead of `sed` and select the last element of the array instead of sort | head.

jesdynf: I assume you're correct and I'm not putting you off when I say I'll study it before the next time I do something like this, but for now if this passes I'll pull it in as it is.

(I was only barely aware bash even had arrays.)

PR #501, opened 2025-11-19
fix(flex): apache log to stdio

kojiromike:
Fixes openemr/openemr#9365

Configures apache on the flex image to log to stdout and stderr instead of logging to /var/log/apache2. This is much easier to work with in a containerized system. You can use `docker compose logs openemr` to see everything, instead of having to exec into the container to extract the logs.

- change the output files to stdio for apache in the flex container

### DIFF
--- a/docker/openemr/7.0.3/openemr.conf
+++ b/docker/openemr/7.0.3/openemr.conf
@@ -18,8 +18,9 @@ Header set X-XSS-Protection "1; mode=block"
 # Narrow document root
 DocumentRoot /var/www/localhost/htdocs/openemr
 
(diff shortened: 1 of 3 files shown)


jesdynf:
I want to keep the option for Apache logs to stay looking like Apache logs. Can we get a Docker envvar that symlinks one or the other of a config fragment during install? You can set your preference (send to Docker) as the default if you want.

kojiromike:
> I want to keep the option for Apache logs to stay looking like Apache logs.

> pl, php-log

I can make this work with both /var/log/apache2/* and stdio. But outputting logs to stdio is more container native, and should be considered the correct way, in containers, going forward. OpenEMR makes a pretty big deal about how docker-forward it is, and this makes things more docker-friendly.

PR #526, opened 2025-12-17
chore: prep dockers for 7.0.4 release

bradymiller:
prep dockers for 7.0.4 release

### DIFF
--- a/.github/workflows/build-705.yml
+++ b/.github/workflows/build-705.yml
@@ -24,7 +24,7 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           context: "{{defaultContext}}:docker/openemr/7.0.5"
-          tags: openemr/openemr:7.0.5, openemr/openemr:dev
+          tags: openemr/openemr:7.0.5, openemr/openemr:dev, openemr/openemr:next
           platforms: linux/amd64,linux/arm64
           push: true
           no-cache: true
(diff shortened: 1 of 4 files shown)


docker/openemr/7.0.4/Dockerfile
```diff
@@ -90,7 +90,7 @@ RUN ln -sf /usr/bin/php${PHP_VERSION_ABBR} /usr/bin/php
 RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin --filename=composer
 
 RUN apk add --no-cache build-base \
-    && git clone https://github.com/openemr/openemr.git --branch rel-704 --depth 1 \
+    && git clone https://github.com/openemr/openemr.git --branch v7_0_4 --depth 1 \
```
kojiromike, line 93: By the way, for maybe a future thing, we can [get these without git](https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives):

```
curl -sSL https://github.com/openemr/openemr/archive/refs/heads/rel-704.tar.gz | tar ...
```

.github/workflows/build-705.yml
stephenwaite, line 12: we probably want to bump this to ubuntu-24.04

docker/openemr/7.0.4/Dockerfile
bradymiller: the new future dockers by @Jmevorach will be doing that :)

PR #529, opened 2025-12-19
Updates 7.0.5 and flex containers materials and adds container benchmarking, test suite tooling, and a new static binary container example

Jmevorach:
Apologies; accidentally wiped out some commits earlier. Now everything is back as it was before and in a single commit!

PR #533, opened 2026-01-01
chore: relabel next planned version from 7.0.5 to 8.0.0

bradymiller:
relabel next planned version from 7.0.5 to 8.0.0

btw, keeping 7.0.5 (along with nightly build) for Jake's ongoing PR (https://github.com/openemr/openemr-devops/pull/529). After bring in that PR, it will stay at 7.0.5 until about 2 weeks from now. At that point will be creating a rel-800 for 8.0.0 release. And at that point will move it from 7.0.5 to 8.0.1.
(this will give adequate time to test Jake's new docker changes since 8.0.0 will be coming out too soon (in 4 weeks) to allow adequate testing)

### DIFF
--- a/docker/openemr/8.0.0/README.md
+++ b/docker/openemr/8.0.0/README.md
@@ -51,7 +51,7 @@ volumes:
   sitevolume: {}
   databasevolume: {}
 ```
-[![Try it!](https://github.com/play-with-docker/stacks/raw/cff22438cb4195ace27f9b15784bbb497047afa7/assets/images/button.png)](http://play-with-docker.com/?stack=https://gist.githubusercontent.com/bradymiller/d285f4ae8d845b3e8a019d1efc44c9ad/raw/8436975f31dc6fea31908f1c533e77ac0e7d2a55/openemr-705-docker-example-docker-compose.yml)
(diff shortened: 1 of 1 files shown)


docker/openemr/8.0.0/upgrade/fsupgrade-8.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
+# Upgrade number 7 for OpenEMR docker
+#  From prior version 7.0.3 (needed for the sql upgrade script).
+priorOpenemrVersion="7.0.3"
```
stephenwaite, line 4: doesn't there have to be an `fsupgrade-9.sh` in order to handle `priorOpenemrVersion="7.0.4"`?

bradymiller:
hi @stephenwaite ,

Generally do this when get closer to release. There are several things that need to do at the same time to support the auto upgrade:
https://www.open-emr.org/wiki/index.php/Steps_for_an_official_release#Prepare_for_docker_upgrade

Rec waiting until after we have a rel-800 branch before doing this (because will then also have a 8.0.1 docker since need to make these changes in both and 8.0.0 and 8.0.1 dockers).

PR #536, opened 2026-01-04
fix: validate github token rate limit is numeric before arithmetic

### LINKED ISSUE #535: openemr.sh arithmetic context confusion with invalid credentials
openemr developer environments are failing to start up with 

```
openemr-1  | trying raw github composer token
openemr-1  |   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
openemr-1  |                                  Dload  Upload   Total   Spent    Left  Speed
100   112    0   112    0     0    263      0 --:--:-- --:--:-- --:--:--   262
openemr-1  | Number of github api requests remaining is null
openemr-1  | Message received from api request is ""Bad credentials""
openemr-1  | ./openemr.sh: line 561: null: unbound variable
```

This is due to a combination of events around https://github.com/openemr/openemr-devops/blob/fb1a80d40aae6b50c002b8a3608d350ced266ada/docker/openemr/flex/openemr.sh#L561.

1. The token in docker-compose.yml is no longer valid. That results in 'Bad credentials'.
2. githubTokenRateLimit gets set to the string 'null'.
3. Because `[[ "${githubTokenRateLimit}" -gt 100 ]]` occurs in arithmetic context, bash see it as `(( $null > 100 ))`.
4. `$null` is not defined, so we get an unbound variable error.

So we need to fix the token, but before we do that we need to ensure that we check if `githubTokenRateLimit` is numeric at all before doing arithmetic on it.

kojiromike:
## Summary
- Add validation to check if the GitHub API rate limit response is empty or non-numeric before attempting arithmetic comparison
- Extract token checking logic into a `try_github_token()` function to avoid duplicating the fix for both raw and encoded token paths

## Problem
When GitHub returns "Bad credentials", `jq '.rate.remaining'` returns the string `null`. The subsequent arithmetic comparison `[[ "${githubTokenRateLimit}" -gt 100 ]]` treats this as a variable reference (`$null`), causing an "unbound variable" error due to `set -u`.

Fixes #535

### DIFF
--- a/docker/openemr/flex/openemr.sh
+++ b/docker/openemr/flex/openemr.sh
@@ -550,54 +550,69 @@ if [[ "${NEED_COMPOSER_BUILD}" = "true" ]] || [[ "${NEED_NPM_BUILD}" = "true" ]]
 
     # Install PHP dependencies if needed
     if [[ "${NEED_COMPOSER_BUILD}" = "true" ]]; then
+        # Try to configure a GitHub token for composer to avoid rate limiting.
+        # Returns 0 on success, 1 on failure.
+        try_github_token() {
+            local token="$1"
+            local token_desc="$2"
+
+            local rate_limit_response
+            local rate_limit
+            local rate_limit_message
+
+            rate_limit_response=$(curl -s -H "Authorization: token ${token}" https://api.github.com/rate_limit)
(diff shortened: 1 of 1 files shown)


bradymiller:
just built new dockers based on this.

PR #549, opened 2026-01-28
docs: change OpenEMR image version to 8.0.1

mschoettle:

Use the correct image in the 8.0.1 README.

- use image tag 8.0.1 for the OpenEMR version
- also removed the version property as it is deprecated in compose

BTW: the link to Play With Docker refers to the wrong gist (v7.0.5)


jesdynf:
Thanks for catching this!

PR #551, opened 2026-02-06
refactor(tools): Prefer `composer phpcs` wrapper

Firehed:
Part of #550, supporting https://github.com/openemr/openemr/pull/10516

#### Short description of what this resolves:
Uses the same phpcs/phpcbf settings as configured in OpenEMR itself

#### Changes proposed in this pull request:
- Uses `composer phpcs` to use OpenEMR's preferred settings
- Same for PHPCBF
- PSR-12 report updated to be closer to that, although it's still off on its own to some extent
- Removes global install of phpcs since it's available through OpenEMR's `require-dev`

PR #552, opened 2026-02-06
refactor(tools): No need to specify extensions in psr12 check anymore

Firehed:
https://github.com/openemr/openemr-devops/pull/551#discussion_r2775620276

### DIFF
--- a/docker/openemr/flex/utilities/devtools
+++ b/docker/openemr/flex/utilities/devtools
@@ -147,7 +147,7 @@ fi
 if [ "$1" = "psr12-src-report" ]; then
     echo "Generating strict PSR12 code styling error report for src directory "
     cd /var/www/localhost/htdocs/openemr
-    composer phpcs -- --standard=PSR12 -d memory_limit=4G --extensions=php,inc -p -n src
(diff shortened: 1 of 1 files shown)


kojiromike (approved):

docker/openemr/flex/utilities/devtools
```diff
@@ -147,7 +147,7 @@ fi
 if [ "$1" = "psr12-src-report" ]; then
     echo "Generating strict PSR12 code styling error report for src directory "
     cd /var/www/localhost/htdocs/openemr
-    composer phpcs -- --standard=PSR12 -d memory_limit=4G --extensions=php,inc -p -n src
+    composer phpcs -- --standard=PSR12 -d memory_limit=4G -p -n src
```
kojiromike, line 150: why do we need to set memory_limit here? Isn't it in the composer line?

Firehed: No, unfortunately `--standard` overrides the `phpcs.xml.dist` file outright.

kojiromike: well isn't that a fine thing.

PR #555, opened 2026-02-08
Harden container CI and improve BATS coverage for 8.0.1/binary/flex

Jmevorach:
This PR improves reliability in both CI and runtime behavior for Docker variants (8.0.1, binary, flex), and makes BATS tests more meaningful by focusing on behavior instead of permissive output matching.

What this changes

1) BATS coverage

- Expands functional BATS coverage across 8.0.1, binary, and flex for startup/config scripts.

2) CI / container test stability

- Fixes container benchmarking log setup by creating test_results/ before tee writes output.
- Runs full container functionality coverage by default; still supports targeted single-test runs via workflow_dispatch.
- Removes fixed host port assumptions from swarm-mode tests to avoid CI conflicts (address already in use).

3) Flex runtime robustness (Redis optionality)

- Makes Redis session configuration conditional on Redis availability.
- Calls wait_for_redis before Redis session setup; if Redis is unavailable, falls back to file sessions instead of failing startup.
- Increases Redis wait tolerance for DNS/startup delays.
- Documents intentional control-flow pattern (if ! wait_for_redis) for shellcheck clarity.

4) CI docker startup correctness

- Ensures test-actions-core starts only intended services (mysql + selected OpenEMR service) during docker compose up, avoiding unintended service startup in production test paths.

Why this matters

- Prevents avoidable CI failures/flakes (ports, missing dirs, service over-start).
- Improves confidence in script behavior validation.
- Allows flex containers to remain healthy when Redis is temporarily unavailable.

### DIFF
--- a/utilities/container_benchmarking/test_suite.sh
+++ b/utilities/container_benchmarking/test_suite.sh
@@ -263,14 +263,6 @@ run_docker_compose() {
     # Unset DOCKER_CONTEXT (Docker CLI variable) to avoid conflicts
     # We don't restore it because we use DOCKERFILE_CONTEXT internally, not DOCKER_CONTEXT
     unset DOCKER_CONTEXT
-
-    # Prune BuildKit cache before builds to prevent snapshot corruption.
-    # After many sequential builds in CI, containerd's snapshotter can lose
-    # parent layers, causing "parent snapshot does not exist" errors.
-    if [[ "${*}" == *"build"* ]]; then
-        docker builder prune -f >/dev/null 2>&1 || true
-    fi
-
     docker compose -p "${project_name}" "$@"
     local exit_code=$?
     return "${exit_code}"
@@ -305,8 +297,6 @@ test_fresh_installation() {
     local test_dir="${TESTS_DIR}/fresh_installation"
     mkdir -p "${test_dir}"
 
-    local docker_context_abs
-    docker_context_abs=$(get_docker_context_abs)
     local flex_env_vars
     flex_env_vars=$(get_flex_env_vars)
 
@@ -332,8 +322,6 @@ services:
       - test-network
 
   openemr:
-    build:
-      context: ${docker_context_abs}
     image: ${IMAGE_TAG}
     depends_on:
       mysql:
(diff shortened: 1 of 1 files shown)


docker/openemr/flex/openemr.sh
```diff
@@ -906,13 +906,23 @@ fi
 if [[ "${REDIS_SERVER}" != "" ]] &&
    [[ ! -f /etc/php-redis-configured ]]; then
 
+    # Only configure PHP Redis when Redis is reachable; otherwise keep file sessions
+    # shellcheck disable=SC2310  # set -e behavior in conditionals is intentional
+    if ! wait_for_redis; then
+        echo "Skipping Redis session config (Redis not available), using file sessions"
+    else
```
bradymiller: hi @Jmevorach , Won't this potentially silently not configure redis and then break downstream orchestration?

docker/openemr/flex/openemr.sh
```diff
@@ -906,13 +906,23 @@ fi
...
     #   Only password set (using redis default user and pertinent password)
     #   NOTE that only username set is not supported (in this case will ignore the username
     #      and use no username and no password set mode)
-    REDIS_PATH="tcp://${REDIS_SERVER}:6379"
+    # REDIS_SERVER may be "host" or "host:port"; avoid double port (tcp://host:6379:6379)
+    if [[ "${REDIS_SERVER}" == *:* ]]; then
+        REDIS_PATH="tcp://${REDIS_SERVER}"
+    else
+        REDIS_PATH="tcp://${REDIS_SERVER}:6379"
+    fi
```
bradymiller, line 929: @Jmevorach , Does this block also need to be in 8.0.1 and binary?

docker/openemr/flex/openemr.sh
Jmevorach: Good catch — you're right, silently falling back to file sessions when REDIS_SERVER is explicitly set could cause subtle issues in orchestrated environments (e.g., sessions not shared across replicas in swarm mode). I've reverted the silent fallback. Now when REDIS_SERVER is set, wait_for_redis is still called to give Redis time to start up and for DNS to resolve, but if it fails, the container exits with an error instead of silently degrading. If the operator configured Redis, we should respect that intent and fail loudly.

docker/openemr/flex/openemr.sh
Jmevorach: No — 8.0.1 and binary already handle this correctly.

Their `configure_redis_sessions()` function uses `IFS=: read -r redis_host redis_port` to parse the host and port separately, then reconstructs the path as `"${redis_host}:${redis_port}"` with a default of 6379. So the "host:port" vs plain "host" format is already handled without risk of double-port. This fix was only needed in flex because flex constructs REDIS_PATH directly as "tcp://${REDIS_SERVER}:6379", which would produce tcp://host:6379:6379 if REDIS_SERVER already contained a port.
 

- See here for 8.0.1's `configure_redis_sessions()` function: https://github.com/openemr/openemr-devops/blob/master/docker/openemr/8.0.1/openemr.sh#L426
- See here for binary's `configure_redis_sessions()` function: https://github.com/openemr/openemr-devops/blob/master/docker/openemr/binary/openemr.sh#L428

PR #565, opened 2026-02-11
fix(flex): only accept 'true' or 'false' for boolean env vars

### LINKED ISSUE #564: Simplify boolean env var handling: only accept 'true'/'false'
## Problem

The current boolean env var handling in `openemr.sh` scripts accepts many different truthy/falsy values (`1`, `yes`, `true`, `on`, `0`, `no`, `false`, `off`, etc.). This causes several problems:

1. **String matching everywhere** - Every check requires `[[ "$VAR" = "true" ]]` or similar
2. **Potential confusion** - `on` and `no` are easy to confuse/typo
3. **Inconsistent behavior** - Different scripts may accept different subsets of values
4. **More code to maintain** - The normalization function adds complexity

## Proposal

Only accept literal `true` and `false` strings for boolean env vars:

```bash
validate_bool() {
    local var_name=$1
    local value=${!var_name}
    case "$value" in
        true|false) ;;
        *) echo "Error: $var_name must be 'true' or 'false', got '$value'" >&2; exit 1 ;;
    esac
}
```

### Benefits

1. **Simpler checks** - Can use `if $VAR; then` directly since `true` and `false` are shell builtins
2. **No ambiguity** - Clear, explicit values only
3. **Fail-fast** - Invalid values caught immediately with helpful error message
4. **Less code** - No normalization logic needed

### Migration

- Update documentation to specify `true`/`false` only
- Add validation that exits with a clear error message for invalid values
- Existing users with `EASY_DEV_MODE=1` or `=yes` will see a clear error telling them to use `=true`

### Affected variables

- `EASY_DEV_MODE`
- `EASY_DEV_MODE_NEW`
- `INSANE_DEV_MODE`
- `PCOV_ON`
- And potentially others that currently accept `yes`/`no`

## Alternative considered

Keep the current normalization approach (accepting many values). Rejected because:
- Adds complexity for marginal user convenience
- Users can easily adapt to `true`/`false`
- Clear error messages make the transition painless

kojiromike:
## Summary

- Replace permissive boolean handling with strict `true`/`false` validation
- Add `validate_bool()` function that exits with clear error message for invalid values
- Simplify all boolean checks to use shell builtins (`if ${VAR}; then`)

## Migration

Users with `EASY_DEV_MODE=1` or `=yes` will see:
```
Error: EASY_DEV_MODE must be 'true' or 'false', got '1'
```

## Test plan

- [ ] Test with valid values (`EASY_DEV_MODE=true`, `EASY_DEV_MODE=false`)
- [ ] Test with invalid values (`EASY_DEV_MODE=yes`) - should fail immediately with clear error
- [ ] Test that container starts correctly with default values

Closes #564

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docker/openemr/compose.yml
+++ b/docker/openemr/compose.yml
@@ -29,7 +29,7 @@ services:
       mysql:
         condition: service_healthy
     environment:
-      MANUAL_SETUP: 'yes'
+      MANUAL_SETUP: 'true'
       MYSQL_HOST: mysql
       MYSQL_PASS: openemr
       MYSQL_ROOT_PASS: root
@@ -63,7 +63,7 @@ services:
     profiles: !override
     - dev
     environment:
-      EMPTY: "yes"
+      EMPTY: 'true'
     volumes:
     - ../../openemr:/var/www/localhost/htdocs/openemr
   openemr-kcov-dev:
--- a/docker/openemr/flex/openemr.sh
+++ b/docker/openemr/flex/openemr.sh
@@ -27,11 +27,11 @@
 #   Optional:
 #     - MYSQL_USER, MYSQL_PASS, MYSQL_DATABASE: Database credentials
 #     - OE_USER, OE_PASS: OpenEMR admin credentials
-#     - EASY_DEV_MODE: 'yes' to skip permission changes (for volume mounts)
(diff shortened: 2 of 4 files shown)


jesdynf (approved):
Looks good to me. I wonder why we set it up this way at first?

PR #566, opened 2026-02-20
fix(easydev): docker pull images

### LINKED ISSUE #567: bug(easydev): openemr-cmd dpi
### Describe the bug

openemr-cmd dpi only prompts to pull images obtained from docker ps and the default (y) to download is ignored

### To Reproduce

1. Add a '....'

### Client configuration

- OpenEMR version:
- Operating system:

stephenwaite:
Fixes #567

openemr-cmd dpi was limited and partially broken

use docker-compose.yml to prompt to download images (defaulted to Y)
-
-
-

### DIFF
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -346,26 +346,24 @@ docker-pull-image(){
         exit 1
     fi
 
-    check_docker_compose_install
+    local services
+    services=$(run_docker_compose -f "${COMPOSE_FILE}" config --services 2>/dev/null)
 
-    local docker_images
-    docker_images=$("${DOCKER_COMPOSE_CMD[@]}" -f "${COMPOSE_FILE}" config --images 2>/dev/null)
-
-    if [[ -z "${docker_images}" ]]; then
-        echo "No images found in ${COMPOSE_FILE}"
+    if [[ -z "${services}" ]]; then
+        echo "No services found in ${COMPOSE_FILE}"
         return
     fi
 
-    local IMAGE RESP
-    while IFS= read -r IMAGE; do
-        [[ -z "${IMAGE}" ]] && continue
-        read -r -e -p "Do you wish to pull ${IMAGE} ? [Y/n] " RESP </dev/tty
+    local SERVICE RESP
+    while IFS= read -r SERVICE; do
+        [[ -z "${SERVICE}" ]] && continue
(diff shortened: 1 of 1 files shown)


utilities/openemr-cmd/openemr-cmd
```diff
@@ -328,20 +330,42 @@ set-webroot(){
...
+    while IFS= read -r IMAGE; do
+        [[ -z "${IMAGE}" ]] && continue
+        read -r -e -p "Do you wish to pull ${IMAGE} ? [Y/n] " RESP </dev/tty
+        if [[ -z "${RESP}" || "${RESP,,}" = y ]]; then
+            docker pull "${IMAGE}"
+        else
+            echo "Skipping ${IMAGE}"
         fi
-    done
+    done <<< "${docker_images}"
```
kojiromike: Would it be simpler to use `docker compose pull` directly instead of parsing images and calling `docker pull` on each?

For pulling all images at once:
```bash
run_docker_compose -f "${COMPOSE_FILE}" pull
```

Or if per-service prompting is desired, iterate over services rather than images:
```bash
local services
services=$(run_docker_compose -f "${COMPOSE_FILE}" config --services)
while IFS= read -r service; do
    [[ -z "${service}" ]] && continue
    read -r -e -p "Pull ${service}? [Y/n] " RESP </dev/tty
    if [[ -z "${RESP}" || "${RESP,,}" = y ]]; then
        run_docker_compose -f "${COMPOSE_FILE}" pull "${service}"
    fi
done <<< "${services}"
```

This avoids the intermediate step of extracting image names — `docker compose pull <service>` already knows which image to pull for each service.

utilities/openemr-cmd/openemr-cmd
```diff
@@ -328,20 +330,42 @@ set-webroot(){
...
+    fi
+
+    local COMPOSE_FILE="${host_openemr_path}/docker/development-easy/docker-compose.yml"
+
+    if [[ ! -f "${COMPOSE_FILE}" ]]; then
+        echo "Error: docker-compose.yml not found at ${COMPOSE_FILE}" >&2
+        exit 1
+    fi
+
+    check_docker_compose_install
```
kojiromike: Minor: The script already has a `run_docker_compose` helper function that handles `check_docker_compose_install` internally. Could use that instead of calling `check_docker_compose_install` directly and then using `"${DOCKER_COMPOSE_CMD[@]}"` separately.

PR #572, opened 2026-03-04
Add Trivy container security scanning, fix ShellCheck targeting, update binary container and harden utility scripts

Jmevorach:
Short description of what this resolves:
- Adds automated container vulnerability scanning for active Docker images, updates the ShellCheck workflow to use .shellcheck-skip marker files instead of hardcoded exclusions, and fixes all ShellCheck warnings in the utilities/ scripts.

Changes proposed in this pull request:
- Add new Trivy Security Scan CI workflow (.github/workflows/trivy.yml) that builds and scans the 8.0.1 and flex Docker images for CRITICAL/HIGH vulnerabilities, using --skip-dirs to scope scanning to OS packages and container-level dependencies rather than application code; includes SARIF upload to the GitHub Security tab
- Replace hardcoded ShellCheck exclusion lists with .shellcheck-skip marker files -- drop a single empty file in any directory to exclude it from scanning, with no workflow YAML changes needed
- Fix all ShellCheck warnings in utilities/openemr-env-installer, utilities/openemr-env-migrator, utilities/openemr-monitor/monitor-installer, and utilities/mariadb-backup-manager/install.sh (quoting variable expansions, direct exit-status consumption, ! a == b to a != b)
- Add Trivy Security Scan badge to README.md


Jmevorach:
@jesdynf Done -- bumped github/codeql-action/upload-sarif from v3 to v4.

jesdynf:
Much obliged! I'm checking on something elsewhere before I pull this in.

.github/workflows/shellcheck.yml
```diff
@@ -3,48 +3,61 @@ on:
...
       - 'kubernetes/kub-up'
-      - 'packages/standard/cfn/buildstacks.bat'  # ¯\_(ツ)_/¯
       - 'utilities/openemr-cmd/openemr-cmd'
       - 'utilities/openemr-cmd/openemr-cmd-h'
       - 'utilities/openemr-env-installer/openemr-env-installer'
       - 'utilities/openemr-env-migrator/openemr-env-migrator'
       - 'utilities/openemr-monitor/monitor-installer'
-      # Exclude everything in obsolete.
+      # Exclude frozen production images and obsolete directories.
+      # These are released images whose scripts must not be modified.
```
kojiromike: The growing exclusion list here is a maintenance concern — each new release adds another `!docker/openemr/X.Y.Z/**` line to five places in the workflow.

The underlying issue is that these directories are being treated as frozen-in-place, but the mechanism for that is a comment and a skip-list rather than something Git actually enforces. If the intent is to pin a released image at a point in time, a tag does that — a directory with a directive not to change it does not.

Separately, 7.0.3 could probably move to `obsolete/` at this point. 7.0.4 and 8.0.0 likely need a bit more maintenance time. But that's all a different PR from this one.

For this PR, the practical question is: does this exclusion approach scale, or should the workflow discover which directories are "active" rather than maintaining a hardcoded skip-list?

.github/workflows/shellcheck.yml
```diff
@@ -124,9 +139,13 @@ jobs:
           # Process each changed file
           for file in "${files_to_check[@]}"; do
             [[ -f "$file" ]] || continue
-            # Skip files in the obsolete directory
+            # Skip files in frozen production or obsolete image directories
+            [[ "$file" = docker/openemr/7.0.3/* ]] && continue
```
kojiromike: This block of `[[ "$file" = ... ]] && continue` checks is growing — could a `case` statement be cleaner here?

```suggestion
            # Skip files in frozen production or obsolete image directories
            case $file in
              docker/openemr/7.0.3/*|docker/openemr/7.0.4/*|docker/openemr/8.0.0/*) continue ;;
              docker/openemr/obsolete/*|docker/obsolete/*) continue ;;
              packages/*) continue ;;
            esac
```

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,86 @@
...
+    name: Trivy (${{ matrix.image }})
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - image: 8.0.1
```
kojiromike: The PR description mentions scanning "the 8.0.1, binary, and flex Docker images" but the matrix only has `8.0.1` and `flex`. Is the `binary` entry intentionally omitted, or should there be a third matrix entry?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,86 @@
...
+            context: docker/openemr/8.0.1
+            build-args: ""
+          - image: flex
+            context: docker/openemr/flex
+            build-args: |
+              ALPINE_VERSION=3.22
+              PHP_VERSION=8.4
```
kojiromike, line 44: These hardcode `ALPINE_VERSION=3.22` and `PHP_VERSION=8.4`. When the flex Dockerfile defaults change, these will need updating in lockstep. Would it be worth adding a comment noting that, or extracting the values from the Dockerfile's ARG defaults?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,86 @@
...
+          scanners: vuln
+          format: table
+          exit-code: 1
+          severity: CRITICAL,HIGH
+          trivyignores: .trivyignore.yaml
+          timeout: 15m
+
+      - name: Run Trivy (SARIF upload)
+        if: always()
+        uses: aquasecurity/trivy-action@0.34.2
```
kojiromike: Both Trivy steps and the SARIF upload use `if: always()`, but the SARIF upload depends on the SARIF file existing (produced by the second Trivy step), and both Trivy steps depend on the image having been built successfully. If the build step fails, `always()` will still attempt to run Trivy against a nonexistent image.

I'd suggest `if: ${{ always() && hashFiles('whatever.sarif') }}` or for the SARIF upload, conditioning on the second Trivy step's outcome would avoid a confusing upload failure.

.trivyignore.yaml
```diff
@@ -0,0 +1,83 @@
+# Trivy ignore rules for OpenEMR Docker images
+# These vulnerabilities come from upstream OpenEMR application dependencies
+# and cannot be fixed in the Dockerfiles. They must be addressed in the
+# OpenEMR application repository (https://github.com/openemr/openemr).
+#
+# MAINTENANCE: When a dependency is updated in the upstream OpenEMR repo,
+# remove its corresponding entry here so Trivy can verify the fix.
```
kojiromike: This file is going to shadow every dependency vulnerability in the OpenEMR application. It's already 83 lines with 20 CVEs, and it will grow with every new advisory.

Since openemr/openemr and openemr/openemr-devops are the same project, there's a double-counting problem: the same vuln gets tracked in the app repo's Dependabot/security advisories *and* manually maintained here. There's no automated mechanism to keep them in sync, so this file will either drift (hiding real issues) or become a constant maintenance chore.

Is there a way to scope the Trivy scan to only Dockerfile-layer and OS-package vulnerabilities — the things the devops repo actually controls — rather than scanning the full application image and ignoring everything it finds? For example, Trivy supports `--skip-dirs` or layer-level scanning. That would let the app repo own its own dependency vulns without this file needing to exist at all.

utilities/mariadb-backup-manager/install.sh
```diff
@@ -169,7 +169,7 @@ chmod 600 restore-client/properties
 installClient () {
     
     (cd backup-client && docker compose -p "${PROJECT}" cp ./ "${SERVICENAME}":"${CLIENTDIRECTORY}")
-    if [[ ! $? == 0 ]]; then
+    if [[ $? != 0 ]]; then
```
kojiromike: The fix from `! $? == 0` to `$? != 0` addresses the ShellCheck warning, but the `$?` anti-pattern remains — the subshell exit status could be consumed directly:

```bash
if ! docker compose --project-directory backup-client -p "${PROJECT}" cp ./ "${SERVICENAME}":"${CLIENTDIRECTORY}"; then
    echo "failure, error writing client to container"
    exit 1
fi
```

kojiromike (changes requested):
Thanks for adding Trivy scanning and cleaning up the ShellCheck workflow. A few structural and correctness questions below.

The biggest concern is scope: the ShellCheck targeting fixes, shell script quoting fixes, and Trivy introduction are independent changes. Would it make sense to split this into at least two PRs — one for the ShellCheck/utility-script fixes and one for Trivy? That way the quoting fixes can land immediately without coupling to the Trivy discussion.

Jmevorach:
@jesdynf 
No problem happy to help anytime! Thanks for reviewing this.

@kojiromike 
Thanks for reviewing this! Appreciate you and Asher's time on this.

That's a fair point about scope. I'm happy to split this if you feel strongly about it, but my rationale for keeping them together was that ShellCheck + Trivy represent the two remaining pieces needed to complete a baseline CI/CD security posture for the devops repo — ShellCheck for static analysis of our shell scripts, Trivy for container-level vulnerability scanning. They felt like a natural unit of work.

That said, if splitting would help get the ShellCheck fixes merged faster, I can pull those into a separate PR. Let me know your preference.

.github/workflows/shellcheck.yml
Jmevorach: Good question. In practice, I don't expect this list to grow much further. All future OpenEMR container images will be based on the 8.0.1 image, which already passes ShellCheck cleanly. The exclusions are only for legacy images (7.0.3, 7.0.4, 8.0.0) that predate the ShellCheck workflow — they won't serve as the base for new releases.

That said, a discovery-based approach (e.g., a marker file like .shellcheck-skip in frozen directories, or an allowlist of active directories instead of a denylist) would be cleaner long-term. I'd be happy to explore that in a follow-up PR.

Agree that moving 7.0.3 to obsolete/ is worth doing separately — that would also naturally shrink this exclusion list.

.github/workflows/trivy.yml
Jmevorach: Intentional. I originally included binary but removed it in a later commit. The binary image is a proof-of-concept that depends on artifacts built from openemr-static-binary-forge in my personal namespace. Including it would significantly expand the Trivy ignore list with CVEs from an older OpenEMR version bundled in those binaries.

I don't want to set a precedent where the project depends on infrastructure in a personal repo. If the community decides to move the binary forge into the OpenEMR org, I'd be happy to bring it under the scan. For now, only the production image (8.0.1) and dev image (flex) are scanned.

I should update the PR description to reflect this — thanks for catching the inconsistency.

.github/workflows/trivy.yml
Jmevorach: Good catch. The intent behind always() on the second Trivy step was to ensure SARIF output is generated even when the first Trivy step fails (exit-code 1 on findings). But you're right that it shouldn't run if the build itself failed.

I'll scope it so the Trivy steps only run when the build succeeds, and the SARIF upload only runs when the SARIF file was actually produced. Something like:

Jmevorach:
```
- name: Run Trivy (SARIF upload)
  if: ${{ success() || steps.trivy-scan.conclusion == 'failure' }}
```

Jmevorach: And for the upload step, gating on hashFiles() as you suggested. I'll push this fix.

utilities/mariadb-backup-manager/install.sh
Jmevorach: Agreed — consuming the exit status directly is cleaner. I intentionally limited my changes to the minimum needed to clear ShellCheck warnings without altering control flow, since these are scripts I can't easily integration-test in CI. But your suggestion is a strict improvement with no behavior change, so I'll adopt it.

.github/workflows/shellcheck.yml
Jmevorach: Nice suggestion — the case statement is definitely more readable and easier to extend. I'll adopt this.

.github/workflows/trivy.yml
Jmevorach: Good callout. The flex Dockerfile uses ARG defaults that can change between releases, so these do need to stay in sync. For now I'll add a comment noting the dependency. Extracting the values programmatically from the Dockerfile's ARG lines is possible (e.g., grep + sed in a prep step) but feels like over-engineering for two values that change infrequently. Happy to revisit if it becomes a maintenance pain.

.trivyignore.yaml
Jmevorach: Fair concern. The ignore file is currently dominated by application-level CVEs because the Alpine base and OS packages happen to be clean right now -- which is actually the success case for container scanning. But you're right that maintaining a manual list of app-level exclusions here creates a sync problem with the app repo's own dependency tracking.

Your --skip-dirs suggestion is a good one. I could scope the scan to exclude the OpenEMR vendor/ and node_modules/ directories, which would eliminate most of these entries and let the app repo own its own dependency vulns. The ignore file would then only contain the rare cases where something falls through the cracks (like the bundled schemaspy.jar which isn't tracked by any package manager).

Happy to explore that in a follow-up -- or I can fold it into this PR if you'd prefer to see it before merging.

PR #604, opened 2026-03-23
fix: switch 8.0.1 Dockerfile from archive to git clone

### LINKED ISSUE #603: Production Docker 8.0.1 test fails: PHPUnit testsuite unit not recognized
The `OpenEMR Production Dockers Test / Production Docker (8.0.1)` job fails because PHPUnit prints its help text and exits 1 when invoked with `--testsuite unit`. This means the `unit` testsuite is not defined in the phpunit.xml bundled in the 8.0.1 production image.

The last successful run on master was 2026-02-20 (PR #555). The test hasn't run on master since then due to path filters, but it now fails when triggered — e.g., on PR #602.

Other production targets (7.0.4, 8.0.0, binary, flex) pass.

kojiromike:
## Summary
- Switch the 8.0.1 Dockerfile's OpenEMR fetch from GitHub archive to `git clone --depth 1`
- OpenEMR's `.gitattributes` now marks `tests/`, `phpunit*.xml`, and other dev files as `export-ignore` (openemr/openemr#11213), which excludes them from archives
- This caused CI to fail because `phpunit.xml` and test directories were missing from the image
- `git clone` ignores export-ignore rules, matching the approach used by the older production Dockerfiles (7.0.4, 8.0.0)

Fixes #603

## Test plan
- [ ] Production Docker (8.0.1) CI job passes

### DIFF
--- a/docker/openemr/8.0.1/Dockerfile
+++ b/docker/openemr/8.0.1/Dockerfile
@@ -147,24 +147,18 @@ RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin -
 # ============================================================================
 # MULTI-STAGE BUILD: OpenEMR Source and Dependencies
 # ============================================================================
-# Stage 1: Fetch OpenEMR source code from GitHub archive
-# Using archive instead of git clone eliminates .git directory overhead
-# Supports tags, branches, or commits via OPENEMR_VERSION build argument
+# Stage 1: Fetch OpenEMR source code via git clone
+# Use git clone instead of GitHub archive because OpenEMR's .gitattributes
+# marks tests/, phpunit*.xml, and other dev files as export-ignore, which
+# excludes them from archives but not from clones.
+# Supports branches or tags via OPENEMR_VERSION build argument.
 # Examples:
-#   - Branch: --build-arg OPENEMR_VERSION=refs/heads/master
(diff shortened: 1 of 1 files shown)


docker/openemr/8.0.1/Dockerfile
```diff
@@ -147,24 +147,18 @@ RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin -
...
+# Supports branches or tags via OPENEMR_VERSION build argument.
 # Examples:
-#   - Branch: --build-arg OPENEMR_VERSION=refs/heads/master
-#   - Tag:    --build-arg OPENEMR_VERSION=refs/tags/7.0.5
-#   - Commit: --build-arg OPENEMR_VERSION=abc123def456
-# Note: GitHub archives extract to openemr-<ref-name> where ref-name is the
-#       last component of the path (e.g., "master" from "refs/heads/master")
-ARG OPENEMR_VERSION=refs/heads/master
+#   - Branch: --build-arg OPENEMR_VERSION=master
+#   - Tag:    --build-arg OPENEMR_VERSION=v7.0.5
```
bradymiller, line 157: @kojiromike , pretty sure this tag is incorrect. Should be `v7_0_5` (or next to be `v8_0_1`). we generally use the branch would be `rel-801`.

PR #613, opened 2026-03-27
Added OPENEMR_GIT arg in Dockerfile to allow pulling from test forks …

### LINKED ISSUE #612: Allow specifying the fork's repo as source repo for building OpenEMR
### Describe the problem

Currently, the docker recipe assumes the image should be built from the OpenEMR repo. 

Converting the git url to an argument in docker recipe can allow the builder to specify a fork's url instead. This can be helpful when building an image for a clinic while awaiting for patchsets to be ready for official release. It also allows for building base images using experimental changes for which you would like to simulate the typical build and release workflow.

luissantosHCIT:
Fixes #612

Allows builder tool to override which fork to pull from when building the OpenEMR image.

- Make the default git repo in Dockerfile an argument defaulting to OpenEMR's official repo.
- Named the new argument `OPENEMR_GIT`
- Added `OPENEMR_GIT` to README


docker/openemr/8.1.0/Dockerfile
```diff
@@ -156,8 +156,9 @@ RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin -
 #   - Branch: --build-arg OPENEMR_VERSION=master
 #   - Tag:    --build-arg OPENEMR_VERSION=v7.0.5
 ARG OPENEMR_VERSION=master
+ARG OPENEMR_GIT=https://github.com/openemr/openemr.git
 FROM base AS openemr-source
-RUN git clone https://github.com/openemr/openemr.git --branch "${OPENEMR_VERSION}" --depth 1 \
+RUN git clone "${OPENEMR_GIT}" --branch "${OPENEMR_VERSION}" --depth 1 \
```
kojiromike, line 159: On the off chance the repo name ends with something other than 'openemr', the directory created won't be named 'openemr'. So an explicit name would be good.

kojiromike (commented):
You may want to pause this to avoid conflicts, or keep an eye on https://github.com/openemr/openemr-devops/issues/614, because the directory will be renamed soon.

docker/openemr/8.1.0/Dockerfile
luissantosHCIT: I'll wait and update this once the other changes are merged. Thank you for the headsup!

One question (I am new to the project), why are the docker directories explicitly versioned as opposed to having a "latest" that gets backed up as the versioned directory prior to updating it further?

Thank you!

stephenwaite: good point @luissantosHCIT, that would be a simpler strategy to have the workflows use next and dev and then stamp them with the version upon release

it'll be worth thinking about moving there after we release 8.1.0 in about a month

PR #643, opened 2026-04-20
style(shell): whole-word quoting, command consolidation, and grep cleanup

kojiromike:

Style follow-up to #459. Related shell-style cleanups applied across the touched files:

1. **Whole-word quoting.** `"${dir}"/documents/era` becomes `"${dir}/documents/era"` — one path, one quoted string. Where the trailing element must stay unquoted (globs, brace expansion), the quote is scoped up to that point (e.g. `"${dir}/path"/*`).
2. **Combine consecutive identical commands** into one multi-arg call (collapsed two `rm -f` in `ssl.sh`, the smarty `rm -fr` pairs in `fsupgrade-{1..9}.sh`, and the prometheus + grafana `mkdir -p` calls in `monitor-installer`).
3. **Group redirects to the same file.** Resolves SC2129 in `openemr.sh`: seven `echo X >> auto_configure.ini` lines (plus a redundant `touch`) become one `{ ...; } > file` brace group, and the two `echo X >> php.ini` opcache-jit lines are wrapped in `{ ...; } >> file`.
4. **`grep` cleanup.**
   - `monitor-installer`: `grep -F` for literal `url` / `smtp`/`to` / port-number matches.
   - `devtoolsLibrary.source` and `flex/utilities/devtools`: drop `-E` where BRE suffices (`^ALTER`, bracket expressions).
   - `flex/utilities/devtools`: replace `echo "$2" | grep -q '[^...]'` identifier validation with POSIX `case $2 in *[!...]*) ...` — one shell, no subprocess.

7.0.4 and 8.0.0 mirrors stay byte-identical so the CI shellcheck content-hash dedup keeps collapsing them. `fsupgrade-8.sh`'s pre-existing header-comment divergence between 7.0.4 and 8.0.0 is left untouched.

No behavior change.

- `docker/openemr/{7.0.4,8.0.0}/openemr.sh` — whole-word quote `"${PHP_VERSION_ABBR}"`-suffixed paths and `fsupgrade-"${c}".sh`; `{ ...; } > auto_configure.ini` brace group; `{ ...; } >> php.ini` brace group.
- `docker/openemr/{7.0.4,8.0.0}/ssl.sh` — whole-word quote `${DOMAIN}` paths; combine `rm -f` pairs.
- `docker/openemr/{7.0.4,8.0.0}/upgrade/fsupgrade-{1..9}.sh` — whole-word quote the `mkdir -p` block, the era/edi/letter_templates/procedure_results migration blocks in fsupgrade-1, and the smarty paths; combine the two smarty `rm -fr` lines.
- `docker/openemr/{7.0.4,8.0.0,8.1.0,8.1.1,binary}/utilities/devtoolsLibrary.source` — drop `-E` from `grep '^ALTER'`.
- `docker/openemr/flex/utilities/devtools` — replace two `grep`-pipeline identifier validators with POSIX `case` statements.
- `utilities/openemr-monitor/monitor-installer` — whole-word quote `${installDir}` paths; collapse mkdirs into one call and drop the redundant `[[ ! -d ]]` guard; `grep -F` for the three literal matches.
- `utilities/openemr-env-installer/openemr-env-installer` — whole-word quote `${brew_installer}` paths.

`shellcheck --include=SC2086,SC2248,SC2046,SC2129` is clean on every touched file.

Out-of-scope follow-ups filed during review: #644, #645, #646, #647, #648.

#### AI disclosure:

### DIFF
--- a/docker/openemr/8.0.0/openemr.sh
+++ b/docker/openemr/8.0.0/openemr.sh
@@ -32,14 +32,15 @@ auto_setup() {
     mkdir "${TMP_FILE_CACHE_LOCATION}"
 
     #create auto_configure.ini to be able to leverage opcache for operations
-    touch auto_configure.ini
-    echo "opcache.enable=1" >> auto_configure.ini
-    echo "opcache.enable_cli=1" >> auto_configure.ini
-    echo "opcache.file_cache=${TMP_FILE_CACHE_LOCATION}" >> auto_configure.ini
-    echo "opcache.file_cache_only=1" >> auto_configure.ini
-    echo "opcache.file_cache_consistency_checks=1" >> auto_configure.ini
-    echo "opcache.enable_file_override=1" >> auto_configure.ini
-    echo "opcache.max_accelerated_files=1000000" >> auto_configure.ini
+    {
+        echo opcache.enable=1
+        echo opcache.enable_cli=1
+        echo "opcache.file_cache=${TMP_FILE_CACHE_LOCATION}"
+        echo opcache.file_cache_only=1
+        echo opcache.file_cache_consistency_checks=1
+        echo opcache.enable_file_override=1
+        echo opcache.max_accelerated_files=1000000
+    } > auto_configure.ini
 
     #run auto_configure
     # shellcheck disable=SC2086 # CONFIGURATION intentionally word-splits into multiple -f flags (see openemr/openemr-devops#539)
@@ -247,7 +248,7 @@ if
         while [ "${c}" -le "${DOCKER_VERSION_ROOT}" ]; do
             if [ "${c}" -gt "${DOCKER_VERSION_SITES}" ] ; then
                 echo "Start: Processing fsupgrade-${c}.sh upgrade script"
-                sh /root/fsupgrade-"${c}".sh
+                sh "/root/fsupgrade-${c}.sh"
                 echo "Completed: Processing fsupgrade-${c}.sh upgrade script"
             fi
             c=$(( c + 1 ))
@@ -270,8 +271,8 @@ if [ "${REDIS_SERVER}" != "" ] &&
     #    version 5.3.7 .
     if [ "${PHPREDIS_BUILD}" != "" ]; then
       apk update
-      apk del --no-cache php"${PHP_VERSION_ABBR}"-redis
-      apk add --no-cache git php"${PHP_VERSION_ABBR}"-dev php"${PHP_VERSION_ABBR}"-pecl-igbinary gcc make g++
+      apk del --no-cache "php${PHP_VERSION_ABBR}-redis"
+      apk add --no-cache git "php${PHP_VERSION_ABBR}-dev" "php${PHP_VERSION_ABBR}-pecl-igbinary" gcc make g++
       mkdir /tmpredis
       cd /tmpredis
       git clone https://github.com/phpredis/phpredis.git
@@ -285,9 +286,9 @@ if [ "${REDIS_SERVER}" != "" ] &&
       ./configure --with-php-config=/usr/bin/php-config83 --enable-redis-igbinary
       make -j "$(nproc --all)"
       make install
-      echo "extension=redis" > /etc/php"${PHP_VERSION_ABBR}"/conf.d/20_redis.ini
+      echo "extension=redis" > "/etc/php${PHP_VERSION_ABBR}/conf.d/20_redis.ini"
       rm -fr /tmpredis/phpredis
-      apk del --no-cache git php"${PHP_VERSION_ABBR}"-dev gcc make g++
+      apk del --no-cache git "php${PHP_VERSION_ABBR}-dev" gcc make g++
       cd /var/www/localhost/htdocs/openemr
     fi
 
@@ -437,14 +438,16 @@ if [ "${XDEBUG_IDE_KEY}" != "" ] ||
    sh xdebug.sh
    #also need to turn off opcache since it can not be turned on with xdebug
    if [ ! -f /etc/php-opcache-jit-configured ]; then
-      echo "opcache.enable=0" >> /etc/php"${PHP_VERSION_ABBR}"/php.ini
+      echo "opcache.enable=0" >> "/etc/php${PHP_VERSION_ABBR}/php.ini"
       touch /etc/php-opcache-jit-configured
    fi
 else
    # Configure opcache jit if Xdebug is not being used (note opcache is already on, so just need to add setting(s) to php.ini that are different from the default setting(s))
    if [ ! -f /etc/php-opcache-jit-configured ]; then
-      echo "opcache.jit=tracing" >> /etc/php"${PHP_VERSION_ABBR}"/php.ini
-      echo "opcache.jit_buffer_size=100M" >> /etc/php"${PHP_VERSION_ABBR}"/php.ini
+      {
+         echo "opcache.jit=tracing"
+         echo "opcache.jit_buffer_size=100M"
+      } >> "/etc/php${PHP_VERSION_ABBR}/php.ini"
       touch /etc/php-opcache-jit-configured
    fi
 fi
--- a/docker/openemr/8.0.0/ssl.sh
+++ b/docker/openemr/8.0.0/ssl.sh
@@ -17,8 +17,7 @@ set -e
 fi
 
 if [ ! -f /etc/ssl/docker-selfsigned-configured ]; then
-    rm -f /etc/ssl/certs/webserver.cert.pem
-    rm -f /etc/ssl/private/webserver.key.pem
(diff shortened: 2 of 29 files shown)


utilities/openemr-monitor/monitor-installer
```diff
@@ -28,44 +28,45 @@ if [[ $# -ne 6 ]]; then
...
+sed -i "s/senderEmail/${senderEmail}/g" "${installDir}/alertmanager.yml"
+sed -i "s/senderUsername/${senderEmail}/g" "${installDir}/alertmanager.yml"
 
 # Input the sender email password
-sed -i "s/senderLoginPassword/${emailPassword}/g" "${installDir}"/alertmanager.yml
+sed -i "s/senderLoginPassword/${emailPassword}/g" "${installDir}/alertmanager.yml"
 
 # Input the receiver email
-sed -i "s/receiverEmail/${receiverEmail}/g" "${installDir}"/alertmanager.yml
+sed -i "s/receiverEmail/${receiverEmail}/g" "${installDir}/alertmanager.yml"
```
kojiromike, line 63: Might it be possible to use a yaml-aware editor like yq here, instead of sed? `sed -i` isn't standard in any case.

utilities/openemr-monitor/monitor-installer
```diff
@@ -28,44 +28,45 @@ if [[ $# -ne 6 ]]; then
...
 echo '******************************Check the Modification******************************'
 echo "Setting in ${installDir}/prometheus/prometheus.yml file."
 grep -E "3001|3002|3003" "${installDir}/prometheus/prometheus.yml"
 echo
 echo "Setting in ${installDir}/grafana/provisioning/datasources/datasource.yml file."
-grep url "${installDir}"/grafana/provisioning/datasources/datasource.yml
+grep url "${installDir}/grafana/provisioning/datasources/datasource.yml"
```
kojiromike: use `grep -F` for a plain-text grep

utilities/openemr-monitor/monitor-installer
```diff
@@ -28,44 +28,45 @@ if [[ $# -ne 6 ]]; then
...
 echo
 echo "Setting in ${installDir}/alertmanager.yml file."
 grep -E 'smtp|to' "${installDir}/alertmanager.yml"
```
kojiromike: There's no good reason to use extended regex here. Use grep -F with multiline here.

utilities/openemr-monitor/monitor-installer
```diff
@@ -28,44 +28,45 @@ if [[ $# -ne 6 ]]; then
...
-curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-monitor/alertmanager.yml           > "${installDir}"/alertmanager.yml
-curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-monitor/docker-compose.yml         > "${installDir}"/docker-compose.yml
-curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-monitor/grafana/provisioning/dashboards/dashboard-193.json > "${installDir}"/grafana/provisioning/dashboards/dashboard-193.json
-curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-monitor/grafana/provisioning/dashboards/dashboard.yml > "${installDir}"/grafana/provisioning/dashboards/dashboard.yml
-curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-monitor/grafana/provisioning/datasources/datasource.yml > "${installDir}"/grafana/provisioning/datasources/datasource.yml
+curl -L https://raw.githubusercontent.com/openemr/openemr-devops/master/utilities/openemr-monitor/prometheus/prometheus.yml  > "${installDir}/prometheus/prometheus.yml"
```
kojiromike, line 39: It's wild to use curl to fetch artifacts from the same repo as this is in. We should have a followup on this.

docker/openemr/8.0.0/ssl.sh
```diff
@@ -44,10 +43,9 @@ if [ "${DOMAIN}" != "" ]; then
 
     # run letsencrypt as a daemon and reference the correct cert
     if [ ! -f /etc/ssl/docker-letsencrypt-configured ]; then
-        rm -f /etc/ssl/certs/webserver.cert.pem
-        rm -f /etc/ssl/private/webserver.key.pem
-        ln -s /etc/letsencrypt/live/"${DOMAIN}"/fullchain.pem /etc/ssl/certs/webserver.cert.pem
-        ln -s /etc/letsencrypt/live/"${DOMAIN}"/privkey.pem /etc/ssl/private/webserver.key.pem
+        rm -f /etc/ssl/certs/webserver.cert.pem /etc/ssl/private/webserver.key.pem
```
kojiromike, line 46: redundancy with the section around line 20 is worth a look

docker/openemr/8.0.0/openemr.sh
```diff
@@ -32,14 +32,15 @@ auto_setup() {
...
-    echo "opcache.max_accelerated_files=1000000" >> auto_configure.ini
+    cat > auto_configure.ini <<EOF
+opcache.enable=1
+opcache.enable_cli=1
+opcache.file_cache=${TMP_FILE_CACHE_LOCATION}
+opcache.file_cache_only=1
+opcache.file_cache_consistency_checks=1
+opcache.enable_file_override=1
+opcache.max_accelerated_files=1000000
+EOF
```
kojiromike: Using a heredoc/nowdoc in php is fine, but due to how indentation works I don't prefer it in shell. Multiline echo is OK, but should be in a compound command.


```suggestion
    {
        echo opcache.enable=1
        echo opcache.enable_cli=1
        echo opcache.file_cache=${TMP_FILE_CACHE_LOCATION}
        echo opcache.file_cache_only=1
        echo opcache.file_cache_consistency_checks=1
        echo opcache.enable_file_override=1
        echo opcache.max_accelerated_files=1000000
    } >> auto_configure.ini
```

docker/openemr/8.0.0/upgrade/fsupgrade-9.sh
```diff
@@ -12,25 +12,24 @@ for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d
...
+        "${dir}/documents/smarty/gacl" \
+        "${dir}/documents/smarty/main" \
+        "${dir}/documents/temp"
     echo "Completed: Ensure have all directories in ${sitename}"
 
     # Clear smarty cache
     echo "Start: Clear smarty cache in ${sitename}"
-    rm -fr "${dir}"/documents/smarty/gacl/*
-    rm -fr "${dir}"/documents/smarty/main/*
+    rm -fr "${dir}/documents/smarty/gacl"/* "${dir}/documents/smarty/main"/*
```
kojiromike, line 32: without dotglob this is a bit questionable. Shouldn't we just delete the directories themselves?

utilities/openemr-monitor/monitor-installer
kojiromike: Agreed this is worth doing but out of scope for this style cleanup. Filed as openemr/openemr-devops#644.

utilities/openemr-monitor/monitor-installer
kojiromike: Applied — changed to `grep -F url` in d670e09.

utilities/openemr-monitor/monitor-installer
kojiromike: Applied — changed to `grep -F -e smtp -e to` in d670e09.

utilities/openemr-monitor/monitor-installer
kojiromike: Good catch — pulling these from the same repo at a pinned ref makes more sense. Out of scope for the style cleanup; filed as openemr/openemr-devops#645.

docker/openemr/8.0.0/ssl.sh
kojiromike: Agreed the whole block looks redundant with what certbot already does. Out of scope here; filed as openemr/openemr-devops#647.

docker/openemr/8.0.0/openemr.sh
kojiromike: Applied your suggested compound-command form in d670e09 (both 7.0.4 and 8.0.0 mirrors).

docker/openemr/8.0.0/upgrade/fsupgrade-9.sh
kojiromike: Good point — the `/*` glob misses dotfiles. Out of scope for quoting/consolidation; filed as openemr/openemr-devops#646.

PR #649, opened 2026-04-20
style(shell): mechanical ShellCheck fixes (SC2164/SC2335/SC2249/SC1010/SC2001)

kojiromike:

First slice of ShellCheck cleanup (follow-up to #459, #643). Pure mechanical fixes — 11 warnings cleared across four files, no judgment calls.

- **ami-build.sh** — SC2249 (getopts `\?)` → `*)`), SC2164 ×3 (`cd … || exit`), SC1010 (quote `echo "ami-build.sh: done"` so shellcheck stops parsing the bare `done` as a loop terminator).
- **ami-configure.sh** — SC2001: `echo "${DVOL}" | sed s/-//` → `${DVOL/-/}`. Same first-only replacement, no subshell, no sed.
- **restore.sh** — SC2249 ×2. The first is the getopts case (cosmetic). The second is a real, quiet bug: the `\?)` branch in `case ${RECOVERYMODE} in` matches the literal string `\?`, not unknown values — so anything besides `local` or `import` silently falls through the whole `case` with `BUCKET` unset, and the next line's `aws s3 cp "s3://${BUCKET}/…"` would try to pull from `s3:///Backup/passphrase.txt`. `*)` actually catches unknown values and exits, which is plainly what the author intended.
- **install.sh** — SC2335 ×2: `! a == b` → `a != b`.

- `packages/standard/ami/ami-build.sh`
- `packages/standard/ami/ami-configure.sh`
- `packages/standard/scripts/restore.sh`
- `utilities/mariadb-backup-manager/install.sh`

Remaining ShellCheck warnings in these files (SC1091, SC2154, SC2312) are in later slices.

#### AI disclosure:

### DIFF
--- a/packages/standard/ami/ami-build.sh
+++ b/packages/standard/ami/ami-build.sh
@@ -13,7 +13,7 @@ while getopts "b:d:" opt; do
     d)
       DOCKERLABEL=${OPTARG}
       ;;
-    \?)
+    *)
       echo "Invalid option: -${opt}" >&2
       exit 1
       ;;
@@ -35,18 +35,13 @@ ln -s /root/aws-cfn-bootstrap-latest/init/ubuntu/cfn-hup /etc/init.d/cfn-hup
 docker pull "openemr/openemr${DOCKERLABEL}"
 
 # get the rest of the devops repo and docker-tools
-cd /root
-if [[ ${REPOBRANCH} = master ]]; then
-  git clone --single-branch https://github.com/openemr/openemr-devops.git && cd openemr-devops/packages/standard
-else
-  git clone --single-branch --branch "${REPOBRANCH}" https://github.com/openemr/openemr-devops.git && cd openemr-devops/packages/standard
-fi
-chmod +x ami/*.sh scripts/*.sh
+cd /root || exit
+git clone --single-branch --branch "${REPOBRANCH}" https://github.com/openemr/openemr-devops.git
+chmod +x openemr-devops/packages/standard/ami/*.sh openemr-devops/packages/standard/scripts/*.sh
 
 # before we get started, we need the AWS cert chain installed
 # we can't grab it later because we can't depend on downloading things
-cd /root
 wget https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
 mv global-bundle.pem mysql-ca
 
-echo ami-build.sh: done
+echo 'ami-build.sh: done'
--- a/packages/standard/ami/ami-configure.sh
+++ b/packages/standard/ami/ami-configure.sh
@@ -9,8 +9,8 @@ source /root/cloud-variables
 
 # prepare the encrypted volume CFN just added
 # this used to be in a weird ready-loop but that doesn't make any sense to me
-DVOL_SERIAL=$(echo "${DVOL}" | sed s/-//)
-DVOL_DEVICE=/dev/$(lsblk -no +SERIAL | grep "${DVOL_SERIAL}" | awk '{print $1}')
+DVOL_SERIAL=${DVOL/-/}
+DVOL_DEVICE=/dev/$(lsblk -no NAME,SERIAL | awk -v s="${DVOL_SERIAL}" '$2 == s {print $1}')
 mkfs -t ext4 "${DVOL_DEVICE}"
 echo "${DVOL_DEVICE}" /mnt/docker ext4 defaults,nofail 0 0 >> /etc/fstab
 mkdir /mnt/docker
--- a/utilities/mariadb-backup-manager/install.sh
+++ b/utilities/mariadb-backup-manager/install.sh
@@ -42,7 +42,7 @@ EOF
 
 identifyProject () {
     if [[ -z "${PROJECT}" ]]; then
-        if [[ ! "$(docker compose ls -q | wc -l)" == 1 ]]; then
+        if [[ "$(docker compose ls -q | wc -l)" != 1 ]]; then
             echo "failure: cannot identify project"
             exit 1
         fi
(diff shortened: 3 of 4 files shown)


packages/standard/ami/ami-build.sh
```diff
@@ -35,18 +35,18 @@ ln -s /root/aws-cfn-bootstrap-latest/init/ubuntu/cfn-hup /etc/init.d/cfn-hup
...
 # get the rest of the devops repo and docker-tools
-cd /root
+cd /root || exit
 if [[ ${REPOBRANCH} = master ]]; then
-  git clone --single-branch https://github.com/openemr/openemr-devops.git && cd openemr-devops/packages/standard
+  git clone --single-branch https://github.com/openemr/openemr-devops.git && cd openemr-devops/packages/standard || exit
 else
-  git clone --single-branch --branch "${REPOBRANCH}" https://github.com/openemr/openemr-devops.git && cd openemr-devops/packages/standard
+  git clone --single-branch --branch "${REPOBRANCH}" https://github.com/openemr/openemr-devops.git && cd openemr-devops/packages/standard || exit
 fi
```
kojiromike: is there any reason to keep this conditional?

packages/standard/ami/ami-build.sh
kojiromike, line 6: hard coding this seems like a bad idea

packages/standard/ami/ami-build.sh
```diff
@@ -35,18 +35,18 @@ ln -s /root/aws-cfn-bootstrap-latest/init/ubuntu/cfn-hup /etc/init.d/cfn-hup
...
 
 # before we get started, we need the AWS cert chain installed
 # we can't grab it later because we can't depend on downloading things
-cd /root
+cd /root || exit
 wget https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
 mv global-bundle.pem mysql-ca
 
-echo ami-build.sh: done
+echo "ami-build.sh: done"
```
kojiromike:
```suggestion
echo 'ami-build.sh: done'
```

ALWAYS use single quotes unless you need a parameter to expand.

packages/standard/ami/ami-build.sh
kojiromike: are the `cd` command needed at all? Could the below chmod just be done with a complete path? Then can we skip the below cd?

packages/standard/ami/ami-configure.sh
```diff
@@ -9,7 +9,7 @@ source /root/cloud-variables
 
 # prepare the encrypted volume CFN just added
 # this used to be in a weird ready-loop but that doesn't make any sense to me
-DVOL_SERIAL=$(echo "${DVOL}" | sed s/-//)
+DVOL_SERIAL=${DVOL/-/}
 DVOL_DEVICE=/dev/$(lsblk -no +SERIAL | grep "${DVOL_SERIAL}" | awk '{print $1}')
```
kojiromike: combine the grep and awk

utilities/mariadb-backup-manager/install.sh
```diff
@@ -169,7 +169,7 @@ chmod 600 restore-client/properties
 installClient () {
     
     (cd backup-client && docker compose -p "${PROJECT}" cp ./ "${SERVICENAME}":"${CLIENTDIRECTORY}")
-    if [[ ! $? == 0 ]]; then
+    if [[ $? != 0 ]]; then
```
kojiromike: combine the lines

packages/standard/ami/ami-build.sh
kojiromike: Good call — the only difference between the two branches was the presence of `--branch`, and `--branch master` works fine against that repo. Collapsed to a single `git clone --single-branch --branch "${REPOBRANCH}" ...` and dropped both `cd` commands; the `chmod` now uses full paths under `openemr-devops/packages/standard/`. Pushed in 7b48cd2.

packages/standard/ami/ami-build.sh
kojiromike: Agreed — the default silently drifts from the current release every time Docker Hub ships a new tag. Filed openemr/openemr-devops#651 to track; keeping it out of this PR since the line isn't changed here and the fix is a design question (fail-fast vs. resolve dynamically vs. tracked version file) rather than mechanical.

packages/standard/ami/ami-build.sh
kojiromike: Fixed — `'ami-build.sh: done'`. Pushed in 7b48cd2.

packages/standard/ami/ami-configure.sh
kojiromike: Combined — `lsblk -no NAME,SERIAL | awk -v s="${DVOL_SERIAL}" '$2 == s {print $1}'`. Constrains the columns (so `+SERIAL` isn't needed) and does an exact match on the serial field rather than a substring grep. Pushed in 7b48cd2.

utilities/mariadb-backup-manager/install.sh
kojiromike: Combined into `if ! (cd backup-client && docker compose ...); then ... exit 1; fi`. Pushed in 7b48cd2.

utilities/mariadb-backup-manager/install.sh
```diff
@@ -168,8 +168,7 @@ chmod 600 restore-client/properties
 
 installClient () {
     
-    (cd backup-client && docker compose -p "${PROJECT}" cp ./ "${SERVICENAME}":"${CLIENTDIRECTORY}")
-    if [[ ! $? == 0 ]]; then
+    if ! (cd backup-client && docker compose -p "${PROJECT}" cp ./ "${SERVICENAME}":"${CLIENTDIRECTORY}"); then
```
kojiromike: This doesn't really need a `cd` command either. There's a flag to docker composer that can replace it, although I'm not sure how you'd have to adjust the cp arguments, too.

kojiromike: Dropped the subshell — `docker compose cp` accepts a host-side path directly, and `backup-client/.` (trailing `/.`) copies the directory's contents rather than the directory itself, matching what `cd backup-client && cp ./ ...` was doing. Pushed in 23e9c1d.

PR #652, opened 2026-04-20
fix(upgrade): preserve dotfiles during fsupgrade-1 migration

### LINKED ISSUE #648: fsupgrade-1: directory migration silently drops dotfiles
`docker/openemr/{7.0.4,8.0.0}/upgrade/fsupgrade-1.sh` migrates four top-level directories into the new `documents/` layout:

- `${dir}/era` → `${dir}/documents/era/`
- `${dir}/edi` → `${dir}/documents/edi/`
- `${dir}/letter_templates` → `${dir}/documents/letter_templates/`
- `${dir}/procedure_results` → `${dir}/documents/procedure_results/`

Each block looks like:

```sh
if [ -d "${dir}/era" ]; then
    if [ "$(ls "${dir}/era")" ]; then
        mv -f "${dir}/era"/* "${dir}/documents/era/"
    fi
    rm -rf "${dir}/era"
fi
```

### Problem

Dotfiles are silently destroyed:

1. `ls "${dir}/era"` — no `-A`, so dotfiles don't appear. The non-empty check returns false if the directory contains only dotfiles.
2. `mv "${dir}/era"/*` — sh's `*` glob skips dotfiles.
3. `rm -rf "${dir}/era"` — runs unconditionally and deletes whatever was left behind.

Net effect: any `.foo` under `era/`, `edi/`, `letter_templates/`, or `procedure_results/` is dropped on upgrade from 5.0.1. ERA = Electronic Remittance Advice (payment/claim files), so this is patient billing data.

### Suggested fix

POSIX `find` enumerates dotfiles and removes the glob/emptiness-test dance:

```sh
if [ -d "${dir}/era" ]; then
    find "${dir}/era/." ! -name . -prune -exec mv -f {} "${dir}/documents/era/" +
    rmdir "${dir}/era"
fi
```

Notes:
- `/.` + `! -name . -prune` enumerates immediate children including dotfiles (POSIX).
- `-exec ... +` batches into one `mv` call.
- `rmdir` replaces `rm -rf` — if anything remained after the move, it fails loudly instead of silently destroying data.

The `letter_templates` block has an extra hoist of `custom_pdf.php` out to `${dir}/` that should remain.

Related: same dotglob class of bug as #646 (smarty cache clear), but the smarty case only loses cache entries. This one loses migrated site data.

kojiromike:

Fixes #648 and clears the remaining ShellCheck warnings in `fsupgrade-1.sh` (10 total: SC2243 x4, SC2312 x4, SC2044 x2). Follow-up slice to #649.

The four directory-migration blocks (era, edi, letter_templates, procedure_results) silently lost dotfiles during the 5.0.1->5.0.2 site layout migration. Three bugs compounded:

1. `[ "$(ls "${dir}/era")" ]` — `ls` skips dotfiles, so a directory containing only dotfiles tested as empty and the `mv` was skipped.
2. `mv -f "${dir}/era"/*` — the glob `*` also skips dotfiles, so any dotfiles present were left behind even when the guard passed.
3. `rm -rf "${dir}/era"` — then destroyed whatever dotfiles remained.

For ERA (Electronic Remittance Advice) and other patient billing data this was silent data loss. Each block now uses the POSIX-safe pattern:

```sh
find "${dir}/era/." ! -name . -prune -exec mv -f {} "${dir}/documents/era/" +
rmdir "${dir}/era"
```

`/.` plus `! -name . -prune` enumerates immediate children including dotfiles. `-exec ... +` batches into one `mv`. `rmdir` replaces `rm -rf` so cleanup fails loudly if anything is left over instead of destroying it. The `custom_pdf.php` hoist in the letter_templates block is preserved unchanged.

The two `for dir in $(find sites/* -maxdepth 0 -type d)` loops are rewritten as simple trailing-slash globs (`for dir in /var/.../sites/*/`), clearing SC2044 and handling whitespace safely.

Two copies of the script remain byte-identical.

- `docker/openemr/7.0.4/upgrade/fsupgrade-1.sh`
- `docker/openemr/8.0.0/upgrade/fsupgrade-1.sh`

#### AI disclosure:

### DIFF
--- a/docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
+++ b/docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
@@ -6,8 +6,9 @@ priorOpenemrVersion="5.0.1"
 echo "Start: Upgrade to docker-version 1"
 
 # Perform codebase upgrade on each directory in sites/
-for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
-    sitename=$(basename "${dir}")
+for dir in /var/www/localhost/htdocs/openemr/sites/*/; do
+    dir="${dir%/}"
+    sitename=${dir##*/}
 
     # Ensure have all directories
     echo "Start: Ensure have all directories in ${sitename}"
@@ -29,33 +30,16 @@ for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d 
 
     # Update new directory structure
     echo "Start: Update new directory structure in ${sitename}"
-    if [ -d "${dir}/era" ]; then
-        if [ "$(ls "${dir}/era")" ]; then
-            mv -f "${dir}/era"/* "${dir}/documents/era/"
-        fi
-        rm -rf "${dir}/era"
+    # letter_templates/custom_pdf.php must be hoisted before the subdir is migrated
+    if [ -f "${dir}/letter_templates/custom_pdf.php" ]; then
+        mv -f "${dir}/letter_templates/custom_pdf.php" "${dir}/"
     fi
-    if [ -d "${dir}/edi" ]; then
-        if [ "$(ls "${dir}/edi")" ]; then
-            mv -f "${dir}/edi"/* "${dir}/documents/edi/"
+    for sub in era edi letter_templates procedure_results; do
+        if [ -d "${dir}/${sub}" ]; then
+            find "${dir}/${sub}/." ! -name . -prune -exec mv -f {} "${dir}/documents/${sub}/" +
+            rmdir "${dir}/${sub}"
         fi
-        rm -rf "${dir}/edi"
-    fi
-    if [ -d "${dir}/letter_templates" ]; then
-        if [ "$(ls "${dir}/letter_templates")" ]; then
-            if [ -f "${dir}/letter_templates/custom_pdf.php" ]; then
-                mv -f "${dir}/letter_templates/custom_pdf.php" "${dir}/"
-            fi
-            mv -f "${dir}/letter_templates"/* "${dir}/documents/letter_templates/"
-        fi
-        rm -rf "${dir}/letter_templates"
-    fi
-    if [ -d "${dir}/procedure_results" ]; then
-        if [ "$(ls "${dir}/procedure_results")" ]; then
-            mv -f "${dir}/procedure_results"/* "${dir}/documents/procedure_results/"
-        fi
-        rm -rf "${dir}/procedure_results"
-    fi
+    done
     echo "Completed: Update new directory structure in ${sitename}"
 
     # Clear smarty cache
@@ -70,8 +54,9 @@ chown -R apache:root /var/www/localhost/htdocs/openemr/sites/
 echo "Completed: Fix permissions"
 
 # Perform database upgrade on each directory in sites/
-for dirdata in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
-    sitename=$(basename "${dirdata}")
+for dirdata in /var/www/localhost/htdocs/openemr/sites/*/; do
+    dirdata="${dirdata%/}"
+    sitename=${dirdata##*/}
 
     # Upgrade database
     echo "Start: Upgrade database for ${sitename} from ${priorOpenemrVersion}"
--- a/docker/openemr/8.0.0/upgrade/fsupgrade-1.sh
+++ b/docker/openemr/8.0.0/upgrade/fsupgrade-1.sh
@@ -6,8 +6,9 @@ priorOpenemrVersion="5.0.1"
 echo "Start: Upgrade to docker-version 1"
 
 # Perform codebase upgrade on each directory in sites/
-for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
-    sitename=$(basename "${dir}")
+for dir in /var/www/localhost/htdocs/openemr/sites/*/; do
+    dir="${dir%/}"
+    sitename=${dir##*/}
 
     # Ensure have all directories
     echo "Start: Ensure have all directories in ${sitename}"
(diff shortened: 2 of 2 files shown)


docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
```diff
@@ -6,7 +6,9 @@ priorOpenemrVersion="5.0.1"
 echo "Start: Upgrade to docker-version 1"
 
 # Perform codebase upgrade on each directory in sites/
-for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
+for dir in /var/www/localhost/htdocs/openemr/sites/*/; do
+    [ -d "${dir}" ] || continue
```
kojiromike: How could this not be a dir at this point? This second check seems unnecessary.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
```diff
@@ -30,31 +32,23 @@ for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d
     # Update new directory structure
     echo "Start: Update new directory structure in ${sitename}"
     if [ -d "${dir}/era" ]; then
```
kojiromike: These could be a loop. Handle custom_pdf.php before entering the loop.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
```diff
@@ -70,7 +64,9 @@ chown -R apache:root /var/www/localhost/htdocs/openemr/sites/
 echo "Completed: Fix permissions"
 
 # Perform database upgrade on each directory in sites/
-for dirdata in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
+for dirdata in /var/www/localhost/htdocs/openemr/sites/*/; do
+    [ -d "${dirdata}" ] || continue
```
kojiromike: redundant check.

docker/openemr/8.0.0/upgrade/fsupgrade-1.sh
```diff
@@ -6,7 +6,9 @@ priorOpenemrVersion="5.0.1"
 echo "Start: Upgrade to docker-version 1"
 
 # Perform codebase upgrade on each directory in sites/
-for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
+for dir in /var/www/localhost/htdocs/openemr/sites/*/; do
```
kojiromike, line 9: Same comments as above.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
kojiromike: Removed — the trailing `/` on `sites/*/` already restricts the glob to directories, and `sites/` always contains at least `default` in this container context, so the fallthrough case the guard was defending against can't happen. Pushed in 342b276.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
kojiromike: Collapsed into a loop over `era edi letter_templates procedure_results`, with the `custom_pdf.php` hoist moved out in front as you suggested. Pushed in 342b276.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
kojiromike: Removed — same reasoning as the first loop's guard. Pushed in 342b276.

docker/openemr/8.0.0/upgrade/fsupgrade-1.sh
kojiromike: All three changes applied to the 8.0.0 copy (it's a straight `cp` of the 7.0.4 file — they stay byte-identical). Pushed in 342b276.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
```diff
@@ -6,7 +6,8 @@ priorOpenemrVersion="5.0.1"
 echo "Start: Upgrade to docker-version 1"
 
 # Perform codebase upgrade on each directory in sites/
-for dir in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
+for dir in /var/www/localhost/htdocs/openemr/sites/*/; do
+    dir="${dir%/}"
     sitename=$(basename "${dir}")
```
kojiromike:
```suggestion
    sitename=${dir##*/}
```

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
```diff
@@ -70,7 +54,8 @@ chown -R apache:root /var/www/localhost/htdocs/openemr/sites/
 echo "Completed: Fix permissions"
 
 # Perform database upgrade on each directory in sites/
-for dirdata in $(find /var/www/localhost/htdocs/openemr/sites/* -maxdepth 0 -type d ); do
+for dirdata in /var/www/localhost/htdocs/openemr/sites/*/; do
+    dirdata="${dirdata%/}"
     sitename=$(basename "${dirdata}")
```
kojiromike: same as above. Why are we doing this loop twice? Should the internals be put into a function and then called within the loop?

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
kojiromike, line 69: Could this be written as…

```suggestion
{
    printf '<?php $_GET["site"] = "%s"; ?>' "${sitename}"
    cat /var/www/localhost/htdocs/openemr/sql_upgrade.php
} | sed -e "/input type='submit'/d" \
        -e "s/!empty(\$_POST\['form_submit'\])/empty(\$_POST\['form_submit'\])/" \
        -e "s/^[   ]*\$form_old_version[   =].*$/\$form_old_version = \"${priorOpenemrVersion}\";/" |
   php
```

?

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
kojiromike: Applied in e1e0e43 — `sitename=${dir##*/}` in both loops, propagated to the byte-identical 8.0.0 copy.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
kojiromike: Same parameter-expansion refactor applied to the second loop in e1e0e43.

On the two-loop structure: the loops can't be merged as-is because `chown -R apache:root /var/www/localhost/htdocs/openemr/sites/` runs once between them, and the DB upgrade needs to see ownership already fixed across every site before any site's `sql_upgrade.php` runs. Merging into one loop would mean either chowning per-site (changes semantics + wasted recursive passes over already-correct sites) or adding a second chown after the loop (belt-and-suspenders at best).

Function extraction for each body is a reasonable separate cleanup, but I'd rather keep that out of this slice — this PR is focused on the dotfile bug + ShellCheck warnings, and a function-extract refactor wants its own test pass against a real 5.0.1 → 5.0.2 fixture. Happy to file a follow-up issue if you want.

docker/openemr/7.0.4/upgrade/fsupgrade-1.sh
kojiromike: Nice — collapsing temp file + three `sed -i` into one pipeline is a real win. Filed openemr/openemr-devops#659 to do it as a separate change, since it needs fixture-backed verification of the upgrade path (in particular, whether `sql_upgrade.php` references `__FILE__`/`__DIR__` — `php` reading from stdin changes those values vs `php -f <path>`, which could silently break the upgrade). Out of scope for this PR, which is scoped to the dotfile fix + ShellCheck cleanup.

PR #653, opened 2026-04-20
style(shell): declare runtime env via stub for ShellCheck (SC1091/SC2154/SC2312)

kojiromike:
## Summary

Follow-up to #649. Resolves the remaining ShellCheck warnings in the three `packages/standard` scripts that source `/root/cloud-variables` at runtime: `ami-configure.sh`, `backup.sh`, and `restore.sh`.

`/root/cloud-variables` is written by CloudFormation UserData at instance boot and is not tracked in the repo, so ShellCheck reports SC1091 (not followable) and SC2154 (variables not assigned) for every variable read from it.

## Changes

- **New `packages/standard/cloud-variables.stub`** declaring the variables CFN injects (`DVOL`, `S3`, `KMS`, `RECOVERYS3`, `RECOVERY_NEWRDS`) with no-op default assignments. The stub is a static declaration only — it is never sourced at runtime; the real values still come from CFN.
- **`# shellcheck source=packages/standard/cloud-variables.stub`** directive added above each `source /root/cloud-variables` line. The path is resolved from the repo root, matching how CI invokes shellcheck. Clears SC1091 and all SC2154.
- **SC2312 hoists**: the remaining `return value masked in pipeline` warnings in `ami-configure.sh` and `backup.sh` are fixed by capturing the producing command's output into a variable, then feeding it to the consumer via herestring. No runtime behavior change.

No `# shellcheck disable` directives are added; the stub plus `source=` directive is the proper fix.

## Verification

```
$ shellcheck --check-sourced --external-sources \
    packages/standard/ami/ami-configure.sh \
    packages/standard/scripts/backup.sh \
    packages/standard/scripts/restore.sh
$ echo $?
0
```

`bash -n` passes on all three scripts.

## Test plan

- [x] `bash -n` clean on all three scripts
- [x] `shellcheck --check-sourced --external-sources` produces no output for the three files
- [ ] CI ShellCheck workflow passes on this PR

### DIFF
--- a/packages/standard/ami/ami-configure.sh
+++ b/packages/standard/ami/ami-configure.sh
@@ -5,12 +5,15 @@ exec > /var/log/openemr-configure.log 2>&1
 cd /root/openemr-devops/packages/standard
 
 # pick up cloud settings
+# shellcheck source=packages/standard/cloud-variables.stub
 source /root/cloud-variables
 
 # prepare the encrypted volume CFN just added
 # this used to be in a weird ready-loop but that doesn't make any sense to me
 DVOL_SERIAL=${DVOL/-/}
-DVOL_DEVICE=/dev/$(lsblk -no NAME,SERIAL | awk -v s="${DVOL_SERIAL}" '$2 == s {print $1}')
+# Split across statements so ShellCheck can see lsblk's exit status (SC2312).
+LSBLK_OUTPUT=$(lsblk -no NAME,SERIAL)
+DVOL_DEVICE=/dev/$(awk -v s="${DVOL_SERIAL}" '$2 == s {print $1}' <<< "${LSBLK_OUTPUT}")
 mkfs -t ext4 "${DVOL_DEVICE}"
 echo "${DVOL_DEVICE}" /mnt/docker ext4 defaults,nofail 0 0 >> /etc/fstab
 mkdir /mnt/docker
--- a/packages/standard/scripts/backup.sh
+++ b/packages/standard/scripts/backup.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 
+# shellcheck source=packages/standard/cloud-variables.stub
 source /root/cloud-variables
 PASSPHRASE=$(aws s3 cp "s3://${S3}/Backup/passphrase.txt" - --sse aws:kms --sse-kms-key-id "${KMS}")
 export PASSPHRASE
-duplicity --full-if-older-than 7D --include "$(docker volume inspect standard_sitevolume | jq -r ".[0].Mountpoint")" --exclude '**' / "boto3+s3://${S3}/Backup"
+VOLUME_INFO=$(docker volume inspect standard_sitevolume)
+MOUNTPOINT=$(jq -r ".[0].Mountpoint" <<< "${VOLUME_INFO}")
+duplicity --full-if-older-than 7D --include "${MOUNTPOINT}" --exclude '**' / "boto3+s3://${S3}/Backup"
 duplicity remove-all-but-n-full 2 --force "boto3+s3://${S3}/Backup"
--- a/packages/standard/cloud-variables.stub
+++ b/packages/standard/cloud-variables.stub
@@ -1,0 +1,9 @@
+# shellcheck shell=sh
+# Declares the variables written into /root/cloud-variables by CloudFormation
+# UserData at instance boot, so ShellCheck can resolve them statically.
+# This file is NOT used at runtime; the real values come from CFN.
+: "${DVOL:=}"           # encrypted EBS volume serial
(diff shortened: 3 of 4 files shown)


packages/standard/ami/ami-configure.sh
```diff
@@ -5,12 +5,15 @@ exec > /var/log/openemr-configure.log 2>&1
...
 
 # pick up cloud settings
+# shellcheck source=packages/standard/cloud-variables.stub
 source /root/cloud-variables
 
 # prepare the encrypted volume CFN just added
 # this used to be in a weird ready-loop but that doesn't make any sense to me
 DVOL_SERIAL=${DVOL/-/}
-DVOL_DEVICE=/dev/$(lsblk -no NAME,SERIAL | awk -v s="${DVOL_SERIAL}" '$2 == s {print $1}')
+LSBLK_OUTPUT=$(lsblk -no NAME,SERIAL)
```
kojiromike, line 15: if LSBLK_OUTPUT isn't used, it can be rolled into the definition of DVOL_DEVICE

packages/standard/ami/ami-configure.sh
kojiromike, line 21: What's the story here?

kojiromike, line 26: dotfiles?

kojiromike, line 55: sed -i is nonstandard and there are a lot of ugly backslashes here. Can this be rewritten with a better parser than sed?

packages/standard/scripts/backup.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
+# shellcheck source=packages/standard/cloud-variables.stub
 source /root/cloud-variables
 PASSPHRASE=$(aws s3 cp "s3://${S3}/Backup/passphrase.txt" - --sse aws:kms --sse-kms-key-id "${KMS}")
 export PASSPHRASE
-duplicity --full-if-older-than 7D --include "$(docker volume inspect standard_sitevolume | jq -r ".[0].Mountpoint")" --exclude '**' / "boto3+s3://${S3}/Backup"
+VOLUME_INFO=$(docker volume inspect standard_sitevolume)
+MOUNTPOINT=$(jq -r ".[0].Mountpoint" <<< "${VOLUME_INFO}")
```
kojiromike, line 8: Are we sure jq exists where this is run? (It ought to.)

packages/standard/ami/ami-configure.sh
kojiromike: Done in bab65e7 — rolled `LSBLK_OUTPUT` back into `DVOL_DEVICE` as a single piped expression. This reintroduces the SC2312 info-level warning (which is why I'd split it in the first place), but readability wins.

packages/standard/ami/ami-configure.sh
kojiromike: Good catch — the `chown 711 /mnt/docker` is suspicious (the comment in the file even wonders whether it was meant to be `chmod`). Filed openemr/openemr-devops#656 to investigate and fix properly; out of scope for this mechanical slice.

kojiromike: Yep — `mv /var/lib/docker/* /mnt/docker` silently skips dotfiles, which is almost certainly wrong here (Docker uses a number of dotted state files/dirs). Filed openemr/openemr-devops#657 to replace with the POSIX dot-preserving idiom; out of scope for this slice.

kojiromike: Agreed — that `sed -i` with escaped `$` and tab literals is ugly and brittle. Filed openemr/openemr-devops#658 to rewrite with a proper PHP-aware approach (likely a small PHP snippet or a dedicated config-edit tool). Out of scope for this mechanical slice.

packages/standard/scripts/backup.sh
kojiromike: Yes — `jq` is installed earlier in the AMI build. `packages/standard/ami/ami-build.sh` runs `dnf install -y … jq …` as part of the base image provisioning, so by the time `backup.sh` runs on the instance (via the daily cron symlink at `/etc/cron.daily/duplicity-backups`), `jq` is guaranteed present.

kojiromike:
Reverted the DVOL_DEVICE rollback (f4942db) — the single-piped form hit SC2312 and the CI ShellCheck job exits 1 on info-level findings. `set -o pipefail` doesn't silence SC2312; the warning is structural (command substitution whose exit status isn't checked), and shellcheck is conservative about `set -e` propagation through command substitution in assignments regardless of pipefail. The split-and-hoist form with a brief comment is the path that keeps CI green without an inline `# shellcheck disable`. Re-pushed after rebasing on latest master.

PR #655, opened 2026-04-21
style(shell): clear ShellCheck warnings in 7.0.4 openemr.sh and ssl.sh

kojiromike:
## Summary

Slice 4 of the openemr-devops ShellCheck cleanup (follow-up to #649, alongside #652 / #653). Resolves all 19 warnings in `docker/openemr/7.0.4/openemr.sh` and both warnings in `docker/openemr/7.0.4/ssl.sh`.

## Changes

- **New `docker/openemr/7.0.4/env.stub`** declares the 14 environment variables both scripts read from the Docker runtime (set by docker-compose `environment:` or `-e` flags): `K8S`, `SWARM_MODE`, `MANUAL_SETUP`, `REDIS_SERVER`, `PHPREDIS_BUILD`, `PHP_VERSION_ABBR`, `REDIS_USERNAME`, `REDIS_PASSWORD`, `REDIS_X509`, `REDIS_TLS`, `XDEBUG_IDE_KEY`, `XDEBUG_ON`, `DOMAIN`, `OPERATOR`. Each is declared with `: "${VAR:=}"` so the stub is a no-op when sourced and never overrides real env values.
- **`# shellcheck source=docker/openemr/7.0.4/env.stub` directive** added to both scripts, followed by `. /root/env.stub 2>/dev/null || true`. The directive path is repo-root-relative (matching how CI invokes shellcheck and matching the pattern from #653); the runtime guard keeps the script working inside the container image, where the stub is not shipped. Clears all SC2154.
- **`openemr.sh` line 80 (SC2188 + SC3020):** replace `{ > FILE ; } &> /dev/null` with `: > FILE 2>/dev/null`. POSIX-compliant no-op write with redirected stderr; same intent (try to create the leader marker, fall back to `AUTHORITY=no` on failure).
- **`openemr.sh` line 256 (SC3037):** replace `echo -n` (POSIX-undefined) with `printf '%s'`.
- **`openemr.sh` line 287 (SC2312):** hoist `nproc --all` into a `jobs` variable so `make -j "${jobs}"` no longer masks nproc's exit status.
- **`openemr.sh` noclobber block:** expand the "atomically test for leadership" comment into a TOCTOU-vs-`O_CREAT|O_EXCL` explanation per review.
- **`openemr.sh` SC2310 ×3** at the retry predicates `while swarm_wait` (line 99) and `while ! auto_setup` (line 241): resolved with targeted `# shellcheck disable=SC2310` plus a comment explaining the intent. The predicates' failure is the loop-termination signal, not a script-fatal condition, so the `set -e` suppression inside the `while` / `!` contexts is the intended behaviour — a structural rewrite would invert the retry-loop semantics.

## Verification

```
$ shellcheck --check-sourced --external-sources \
    docker/openemr/7.0.4/openemr.sh \
    docker/openemr/7.0.4/ssl.sh
```

is clean (exit 0).

## Test plan

- [x] `bash -n` clean on both scripts
- [x] `shellcheck --check-sourced --external-sources` clean on both scripts
- [ ] CI ShellCheck workflow passes on this PR

### DIFF
--- a/docker/openemr/7.0.4/openemr.sh
+++ b/docker/openemr/7.0.4/openemr.sh
@@ -12,6 +12,18 @@ set -e
 # shellcheck source=SCRIPTDIR/utilities/devtoolsLibrary.source
 . /root/devtoolsLibrary.source
 
+# Environment variables used by this script are supplied by the Docker runtime
+# (docker-compose `environment:` or `-e` flags). The env.stub file declares
+# them for ShellCheck's benefit using `: "${VAR:=}"` assignments that leave
+# real runtime values untouched. The stub is not shipped into the container
+# image; the `if false` keeps the `.` statically visible to ShellCheck's
+# source-follower without ever running it — BusyBox ash treats `.` as a
+# special builtin and exits the shell on file-not-found even with `|| true`.
+if false; then
+    # shellcheck source=docker/openemr/7.0.4/env.stub
+    . /root/env.stub
+fi
+
 swarm_wait() {
     if [ ! -f /var/www/localhost/htdocs/openemr/sites/docker-completed ]; then
         # true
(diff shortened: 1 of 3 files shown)


docker/openemr/7.0.4/openemr.sh
```diff
@@ -77,7 +85,7 @@ fi
 if [ "${SWARM_MODE}" = "yes" ]; then
     # atomically test for leadership
     set -o noclobber
-    { > /var/www/localhost/htdocs/openemr/sites/docker-leader ; } &> /dev/null || AUTHORITY=no
+    : > /var/www/localhost/htdocs/openemr/sites/docker-leader 2>/dev/null || AUTHORITY=no
     set +o noclobber
```
kojiromike, line 99: What if we just check if the file exists instead of setting noclobber and unsetting it?

kojiromike: The comment on line 86 — "atomically test for leadership" — is the key constraint: this runs in swarm mode across multiple racing containers, all of which execute this block at roughly the same instant, and exactly one must become the leader.

`[ -f file ]` followed by creating it has a TOCTOU race: two (or more) containers can each see "file doesn't exist" in the check, both then create it, and both conclude they are the leader. That's exactly the split-brain the existing code is written to avoid.

`set -o noclobber` + `: > file 2>/dev/null` delegates the race to the kernel's `open(O_CREAT|O_EXCL)` — the file is created atomically by whichever container the kernel serializes first, and every other container's redirect fails with `EEXIST`, which the `|| AUTHORITY=no` branch catches. One winner, by construction.

So I'd like to leave this alone — the noclobber toggle is doing real work, not style. Happy to add an inline comment explaining *why* noclobber instead of `[ -f ]` if that would help future readers; just say the word.

kojiromike: Added the TOCTOU explanation in b60007f — noted the race that motivates `noclobber` and how `open(O_CREAT|O_EXCL)` serializes the creation.

PR #663, opened 2026-04-27
fix(flex): defer swarm follower wait until after runtime build

### LINKED ISSUE #661: CI flake: Production Docker (flex) sometimes fails with "application not healthy after 10m0s"
## What

The flex production docker job (`build / Production Docker (flex)`) intermittently fails with:

```
application not healthy after 10m0s
##[error]Process completed with exit code 1.
```

Container logs show mysql healthy and openemr started, but the openemr container's healthcheck never passes within the 600-second wait window.

## Where it comes from

`.github/actions/test-actions-core/action.yml` runs:

```yaml
- name: Run the containers
  run: docker compose up --detach --wait --wait-timeout 600 mysql "\${OPENEMR_SERVICE_NAME}"
```

`docker compose up --wait` polls the service healthcheck and exits non-zero with that message when the timeout elapses.

## Recent occurrence

Run https://github.com/openemr/openemr-devops/actions/runs/24732488650 on master (commit 959f246, the merge of #660). Re-running the failed job alone passed without changes, confirming flake.

- #660 only touched `docker/openemr/8.0.0/*` — did not modify the flex container.
- The flex job passed on the PR CI for #660 and on the master push CI for the merge immediately prior (#655, commit 84ca0c3).

## Why it likely flakes

The flex image runs `composer install` at container start. Under CI runner load (cold cache, slow mirror, contention), that can push total boot past 10 min. The wait-timeout is a budget, not a correctness check.

## Suggested mitigations (pick one)

- **Raise the `--wait-timeout`.** 20m would absorb most composer-install variance while still catching real hangs.
- **Warm the image.** Move composer install into the image build (cached layer) rather than container startup, so healthcheck measures runtime boot only.
- **Retry the failed step once** before failing the job. Cheapest fix, doesn't address root cause but silences the flake.
- **Split the healthcheck into cheap (apache up) + slow (installer done) signals**, so compose's `--wait` succeeds quickly while a separate step verifies post-install state.

## Priority

Low — flake is rare and resolved by a single re-run. Open to document the pattern and the mitigation options so the next person who hits it doesn't have to re-derive the diagnosis.

Jmevorach:
Flex containers in SWARM_MODE=start the follower alongside the leader. The follower was blocking on the leader before cloning/building OpenEMR, so its Docker health start_period was consumed during the wait and then again during npm/composer—often exceeding the health check window and failing CI.

- Introduce SWARM_WAIT_DEFERRED: followers skip wait_for_swarm_completion until after the local flex build block, then run wait + prepare_swarm_leader.
- Split wait vs leader prep into wait_for_swarm_completion and prepare_swarm_leader for a clear call order.
- Fix leader takeover checks: try_become_leader always returns 0, so gate promotion on AUTHORITY=yes instead of treating the return code as success.

Tests:
- Bats: assert_script_syntax on flex openemr.sh; smoke checks for swarm deferral.

Validated with: bats tests/bats/flex/openemr.bats and ./utilities/container_benchmarking/test_functionality.sh flex --test swarm_mode

Closes #661

### DIFF
--- a/docker/openemr/flex/openemr.sh
+++ b/docker/openemr/flex/openemr.sh
@@ -288,6 +288,7 @@ is_configured() {
 
 AUTHORITY=yes
 OPERATOR=yes
+SWARM_WAIT_DEFERRED=no
 
 # Kubernetes-specific role assignment
 if [[ "${K8S}" = "admin" ]]; then
@@ -377,17 +378,8 @@ update_leader_heartbeat() {
     fi
 }
 
-# Handles swarm mode coordination: leader election and follower waiting.
-handle_swarm_mode() {
-    # Skip coordination if swarm mode isn't enabled
-    if [[ "${SWARM_MODE}" != "yes" ]]; then
-        return 0
-    fi
-
-    # Try to become the leader
-    try_become_leader
-
-    # If we're a follower, wait for the leader to finish setup
+# Waits for the swarm leader to finish setup before reading shared configuration.
+wait_for_swarm_completion() {
     if [[ "${AUTHORITY}" = "no" && ! -f "${OE_ROOT}/sites/docker-completed" ]]; then
         echo "Waiting for docker-leader to finish configuration (with timeout-based recovery)..."
         local -i max_wait_time="${LEADER_WAIT_TIMEOUT:-600}"  # Default: 10 minutes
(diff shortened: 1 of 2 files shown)


Jmevorach:
@kojiromike closes https://github.com/openemr/openemr-devops/issues/661

kojiromike (approved):
Clean refactor with a real bug fix (try_become_leader return-code gating) and a sensible deferred-wait pattern that lets flex followers parallelize their local build with the leader's shared-volume setup. CI green across the matrix. New assert_script_syntax test is a nice safety net.

PR #687, opened 2026-05-04
ci: rebuild flex 3.23 image nightly

### LINKED ISSUE #686: Auto-build openemr/openemr:flex on a nightly schedule
## Problem

The default `openemr/openemr:flex` tag is published only on manual `workflow_dispatch` runs of `.github/workflows/build-323.yml` (which currently builds the Alpine 3.23 / PHP 8.5 default flex image).

Nothing automatic refreshes it when commits land on either `openemr-devops@master` or `openemr/openemr@master`. As a result, fixes can sit unpublished for weeks and `docker pull openemr/openemr:flex` returns a stale image.

A concrete recent example: `fix(flex): use sql_upgrade.php CLI mode in upgradeOpenEMR` (#666) merged on 2026-05-01, but as of today the published `openemr/openemr:flex` digest is still the 2026-04-07 build, so `openemr-cmd dev-reset-install-demodata` continues to fail in a fresh container until the workflow is dispatched manually.

## Why this matters

The flex image is the development image users get from `docker/development-easy/docker-compose.yml` upstream. When devops fixes don't reach it, every developer running `openemr-cmd drid` (or anything else that exercises the bundled `devtoolsLibrary.source`) hits the old bug — even with a fully up-to-date `openemr/openemr` checkout, since the image bakes both repos' contents.

## Suggested direction

Add a `schedule` trigger (and keep `workflow_dispatch`) to `.github/workflows/build-323.yml`, mirroring the pattern already in use for `build-edge.yml`:

```yaml
on:
  workflow_dispatch:
  schedule:
    - cron: '0 2 * * *'
```

A nightly rebuild keeps the `flex` floating tag within ~24h of the latest devops + openemr master, at the cost of one extra image build per day (and modest registry churn).

If preferred, a `workflow_run` / `repository_dispatch` trigger from the openemr/openemr master branch is also viable, but cron is simpler and matches the existing edge-build precedent.

When a future Alpine version becomes the new default, the same pattern should move with it (whichever workflow holds `is_default_flex: true`).

kojiromike:
## Summary

Adds a `schedule` trigger to `.github/workflows/build-323.yml` so the default `openemr/openemr:flex` tag refreshes nightly at 2 AM UTC, matching the existing pattern in `build-edge.yml`, `build-810.yml`, and `build-811.yml`.

Without this, fixes that land on `master` (devops or upstream openemr) can sit in the unpublished `openemr/openemr:flex` image for weeks until someone manually dispatches the workflow. The motivating example: #666 merged on 2026-05-01, but the published `flex` digest is still the 2026-04-07 build, so `openemr-cmd dev-reset-install-demodata` continues to fail in fresh containers.

When a future Alpine version takes over `is_default_flex: true`, this trigger should move with it.

Closes #686

## Test plan

- [ ] Verify actionlint passes
- [ ] After merge, confirm a scheduled run kicks off the next 2 AM UTC and publishes a fresh `openemr/openemr:flex` digest

### DIFF
--- a/.github/workflows/build-323.yml
+++ b/.github/workflows/build-323.yml
@@ -17,6 +17,8 @@ name: Flex 3.23 Dockers Build
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '0 2 * * *' # run at 2 AM UTC
 
 jobs:
   build:


bradymiller:
hi @kojiromike and @stephenwaite ,
This makes good sense. I wonder if we should just do this for all the flex build scripts (currently just this one and the 3.22 one). Not only will they stay updated, but we will know if they break. Would that make sense?

PR #693, opened 2026-05-06
feat(openemr-cmd): add 'worktree set-env' to switch a worktree's env

bradymiller:
## Summary
- Adds `openemr-cmd worktree set-env <branch> <easy|easy-light|easy-redis>` so an existing worktree can be moved between env variants without removing and re-adding the worktree.
- Reuses the existing `wt_write_env` / `wt_write_override` generators — no new file-generation logic — and just swaps the `env` field in `.worktrees.json` (offset and dir are preserved).
- Requires the worktree's stack to be in `none` state (i.e. `worktree down` first) before switching, so the user explicitly chooses the data-handling path (`worktree down` vs `worktree down --keep-volumes`) rather than the script silently discarding or stranding old-env volumes.
- Bumps `VERSION` to `1.0.29` and lists the new subcommand in both the `worktree` usage block and the top-level `-h` help.

## Test plan
- [ ] `openemr-cmd worktree` and `openemr-cmd -h` both list the new `set-env` subcommand
- [ ] `worktree set-env <branch> <invalid-env>` is rejected by `wt_validate_env`
- [ ] `worktree set-env <unknown-branch> easy` errors with "No worktree found for branch"
- [ ] `worktree set-env <branch> <same-env>` is a no-op and prints "already on env"
- [ ] `worktree set-env <branch> <new-env>` while the stack is `running` or `stopped` is refused with a hint to run `worktree down` first
- [ ] `worktree down <branch>` then `worktree set-env <branch> <new-env>` updates `.worktrees.json` and rewrites the new env's `.env` and `docker-compose.override.yml`; `worktree up <branch>` then comes up cleanly with the new env's services
- [ ] Switching back to a previously-used env produces the expected fresh `.env` and override (no stale content from the prior visit)
- [ ] `worktree list` reflects the new env after the switch

### DIFF
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -18,7 +18,7 @@ set -euo pipefail
 #################################################################################################
 
 # Increment the version when modify script
-VERSION="1.0.28"
+VERSION="1.0.29"
 
 # ============================================================================
 # WORKTREE STATE MANAGEMENT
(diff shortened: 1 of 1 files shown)


bradymiller:
testing well and bringing in

PR #718, opened 2026-05-13
feat(openemr-cmd): wrap pre-commit tooling, add prek-in-container dispatch

bradymiller:
## Summary

Adds in-container access to every pre-commit hook from OpenEMR's `.pre-commit-config.yaml` so developers can `git commit` from the host **without installing PHP, Node, Python, codespell, pre-commit, or actionlint locally**. The openemr flex container becomes the execution environment; openemr-cmd is the bridge.

- **openemr-cmd wrappers for each pre-commit composer script**: `cps` (codespell), `ccc` (conventional-commits-check), `crc` (require-checker), `cv` (composer-validate), `cn` (composer-normalize), `cnf` (composer-normalize-fix), `cck` (composer-checks), `cq` (code-quality). First five wired into `clean-sweep` so the existing aggregate stays comprehensive.
- **`openemr-cmd prek <args>`**: passthrough to in-container `pre-commit`. Substitutes `actionlint-docker` → `actionlint-system` at invocation time (DinD isn't available from inside the container); the repo's `.pre-commit-config.yaml` is untouched and host-side prek/pre-commit usage continues to work unchanged. cwd-aware container resolution (`wt_resolve_container_from_cwd`) so a commit fired from worktree `foo` always dispatches to `foo`'s container.
- **`openemr-cmd pi` / `pu`** (`prek-install` / `prek-uninstall`): manage host-side `.git/hooks/{pre-commit,commit-msg}` shims. One install on the primary repo covers every linked worktree (current and future) via gitcommondir. Refuses to clobber non-managed hooks without `--force`. Runtime safeguard rejects invocation via `-d <container>` or `worktree exec` with a clear error.
- **Container packages relocated**: `chromium`, `chromium-chromedriver`, `py3-codespell`, plus the new `pre-commit` and `actionlint`, now installed at top of `openemr.sh` under `DEVELOPER_TOOLS=yes` gate (idempotent via `apk info -e` guards). Previously inside `NEED_COMPOSER_BUILD=true`, which meant they didn't survive `docker compose down --keep-volumes && up` (vendor preserved → block skipped → packages missing on the recreated container).
- Bumps `openemr-cmd` VERSION to 1.0.31. Fixes a stale `openemr-cmd-h` filter line count (was 9, now 23).

## End-to-end workflow this unlocks

For a contributor with only Docker installed on their host:

```sh
openemr-cmd worktree add my-feature -b              # one-time per feature
openemr-cmd pi                                       # one-time per clone
# edit files
git add path/to/file.php
git commit -m "feat: ..."
# → host hook fires → openemr-cmd prek run --hook-stage pre-commit
# → cwd-aware resolution dispatches to my-feature's container
# → pre-commit runs inside container against staged files
# → commit aborts on any failure, succeeds otherwise
```

Manual invocation paths remain:
- `openemr-cmd prek run --all-files` — full-codebase check
- `openemr-cmd cq` — full composer code-quality suite (direct composer path, bypasses prek)
- `openemr-cmd worktree exec <branch> prek run --all-files` — target a specific worktree from anywhere

### `openemr-cmd` smoke checks
- [x] `openemr-cmd --version` reports 1.0.31
- [x] `openemr-cmd --help` shows the `pre-commit-management:` section with `prek`, `pi`/`prek-install`, `pu`/`prek-uninstall` entries, plus the eight new short-form wrappers under `php-management`
- [x] `openemr-cmd-h php` displays all 23 php-management entries (was truncated before due to stale filter count)

### Per-tool wrappers (run inside any DEVELOPER_TOOLS=yes container; `openemr-cmd` from host)
- [x] `openemr-cmd cps` → runs codespell, no crash on warnings
- [x] `openemr-cmd cv` → reports composer.json validation
- [x] `openemr-cmd cn` → dry-run normalization, reports diff or "already normalized"
- [x] `openemr-cmd crc` → composer-require-checker output
- [x] `openemr-cmd cq` → runs the full composer code-quality suite

### prek dispatcher
- [x] `openemr-cmd prek --version` reports the in-container pre-commit version
- [x] `openemr-cmd prek run --all-files` executes the full hook suite; actionlint runs without DinD
- [x] `openemr-cmd prek run phpstan` runs the single hook
- [x] Inside the openemr container, `/tmp/openemr-cmd-pre-commit-config.yaml` shows `actionlint-system` (not `-docker`)
- [x] From inside worktree `foo`, `openemr-cmd prek run` dispatches to `foo`'s container (verify via `docker ps` while it runs)

### Hook installation
- [x] `openemr-cmd pi` from the primary repo writes `pre-commit` and `commit-msg` shims to `<primary>/.git/hooks/`
- [x] Both shim files contain the `__openemr_cmd_prek_hook__` marker
- [x] `git commit` from a worktree fires the shim and dispatches into that worktree's container
- [x] `git commit` from the primary fires the shim and dispatches into the primary's container
- [x] `openemr-cmd pi` refuses to overwrite a non-managed pre-existing hook; `pi --force` backs it up to `*.bak.<timestamp>` and installs
- [x] `openemr-cmd pu` removes only marker-bearing hooks; skips non-managed ones with a notice

### Runtime safeguards
- [x] `openemr-cmd -d <any-container> pi` exits 1 with "host-only command" error (does not enter container)
- [x] `openemr-cmd worktree exec <branch> pi` produces the same error transitively
- [x] Same for `pu`/`prek-uninstall`

### Container package persistence (the NEED_COMPOSER_BUILD fix)
- [x] After `openemr-cmd worktree down --keep-volumes <branch> && openemr-cmd worktree up <branch>`, all of `chromium-browser`, `codespell`, `pre-commit`, and `actionlint` resolve to installed binaries inside the new container
- [x] On subsequent container starts with packages already present, no apk install runs (verify the `apk info -e` short-circuit)

### Compatibility
- [ ] Host-based `prek install` followed by `git commit` still works for users who haven't run `openemr-cmd pi` (no change to host workflow)
- [x] CI continues to pass; in particular composer code-quality and shellcheck don't regress

### DIFF
--- a/docker/openemr/flex/utilities/devtools
+++ b/docker/openemr/flex/utilities/devtools
@@ -27,6 +27,14 @@
 #   rector-dry-run              - Preview PHP code modernization changes
 #   rector-process              - Apply PHP code modernization changes
 #   phpstan                     - Run PHPStan static analysis
+#   codespell                   - Run codespell (composer codespell)
+#   conventional-commits-check  - Validate conventional commit messages
+#   require-checker             - Run composer-require-checker
+#   composer-validate           - Validate composer.json (strict)
+#   composer-normalize          - Check composer.json normalization (dry-run)
+#   composer-normalize-fix      - Apply composer.json normalization
+#   composer-checks             - Run composer "checks" (validate + normalize dry-run)
+#   code-quality                - Run full composer "code-quality" suite
 #
 # BUILD & ASSETS:
 #   build-themes            - Build OpenEMR CSS themes
@@ -210,6 +218,54 @@ if [ "$1" = "phpstan-generate" ]; then
     echo "PHPStan regenerate baseline"
     cd /var/www/localhost/htdocs/openemr
     composer phpstan-baseline
+fi
+
+if [ "$1" = "codespell" ] || [ "$1" = "clean-sweep" ]; then
+    echo "Running codespell"
+    cd /var/www/localhost/htdocs/openemr
+    composer codespell
+fi
+
+if [ "$1" = "conventional-commits-check" ] || [ "$1" = "clean-sweep" ]; then
+    echo "Validating conventional commit messages"
+    cd /var/www/localhost/htdocs/openemr
+    composer conventional-commits:check
+fi
+
+if [ "$1" = "require-checker" ] || [ "$1" = "clean-sweep" ]; then
+    echo "Running composer-require-checker"
+    cd /var/www/localhost/htdocs/openemr
+    composer require-checker
+fi
+
+if [ "$1" = "composer-validate" ] || [ "$1" = "clean-sweep" ]; then
+    echo "Validating composer.json (strict)"
+    cd /var/www/localhost/htdocs/openemr
+    composer validate --strict
+fi
+
+if [ "$1" = "composer-normalize" ] || [ "$1" = "clean-sweep" ]; then
+    echo "Checking composer.json normalization"
+    cd /var/www/localhost/htdocs/openemr
+    composer normalize --dry-run
+fi
+
+if [ "$1" = "composer-normalize-fix" ]; then
+    echo "Applying composer.json normalization"
+    cd /var/www/localhost/htdocs/openemr
+    composer normalize
+fi
+
+if [ "$1" = "composer-checks" ]; then
+    echo "Running composer checks (validate + normalize dry-run)"
+    cd /var/www/localhost/htdocs/openemr
+    composer checks
+fi
+
+if [ "$1" = "code-quality" ]; then
+    echo "Running full composer code-quality suite"
+    cd /var/www/localhost/htdocs/openemr
+    composer code-quality
 fi
 
 # ============================================================================
--- a/docker/openemr/flex/utilities/devtools
+++ b/docker/openemr/flex/utilities/devtools
@@ -203,7 +203,10 @@ do_lint_javascript_fix() {
 
 do_php_parserror() {
     echo "Generating PHP parse errors"
-    run_at_openemr_root sh -c 'find . -type f \( -name "*.php" -or -name "*.inc" \) \( -not -path "./vendor/*" -and -not -path "./node_modules/*" -and -not -path "./ccdaservice/node_modules}/*" \) -exec php -d error_reporting=32767 -l {} \; 2>&1 >&- | grep "^"'
+    # Fix pre-existing typo: './ccdaservice/node_modules}/*' had a stray '}'
+    # so the exclusion silently never matched, causing php -l to be run on
+    # every .php file under ccdaservice/node_modules. Now correctly excluded.
+    run_at_openemr_root sh -c 'find . -type f \( -name "*.php" -or -name "*.inc" \) \( -not -path "./vendor/*" -and -not -path "./node_modules/*" -and -not -path "./ccdaservice/node_modules/*" \) -exec php -d error_reporting=32767 -l {} \; 2>&1 >&- | grep "^"'
 }
 
 do_rector_dry_run() {
--- a/docker/openemr/flex/utilities/devtools
+++ b/docker/openemr/flex/utilities/devtools
@@ -122,6 +122,15 @@
 
 source /root/devtoolsLibrary.source
 
+# Run `composer <args>` from the openemr web root. Lets handlers say what
+# they're doing in one line instead of the cd-then-composer two-step.
+# `cd || exit 1` aborts the script if the webroot is missing rather than
+# silently invoking composer from a wrong cwd (devtools has no `set -e`).
+run_composer_at_root() {
+    cd /var/www/localhost/htdocs/openemr || exit 1
+    composer "$@"
+}
+
 # ============================================================================
 # LOGGING COMMANDS
 # ============================================================================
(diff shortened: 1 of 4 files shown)


docker/openemr/flex/utilities/devtools
```diff
@@ -212,6 +220,54 @@ if [ "$1" = "phpstan-generate" ]; then
     composer phpstan-baseline
 fi
 
+if [ "$1" = "codespell" ] || [ "$1" = "clean-sweep" ]; then
```
kojiromike: Suggest case/esac for this section, instead of if. There also really ought to be a way to avoid duplicating that cd. A shell function perhaps?

utilities/openemr-cmd/openemr-cmd
```diff
@@ -162,6 +162,42 @@ wt_validate_dir_safe() {
...
+# wrong container when multiple worktree stacks are running concurrently.
+wt_resolve_container_from_cwd() {
+    local toplevel toplevel_real branch slug
+    toplevel=$(git rev-parse --show-toplevel 2>/dev/null) || return 0
+    toplevel_real=$(realpath -m "${toplevel}" 2>/dev/null) || return 0
+    [[ -f "${WT_STATE_FILE}" ]] || return 0
+
+    while IFS=$'\t' read -r b d; do
+        [[ -z "${b}" ]] && continue
+        local d_real
```
kojiromike: Can hoist this declaration

utilities/openemr-cmd/openemr-cmd
```diff
@@ -162,6 +162,42 @@ wt_validate_dir_safe() {
...
+        d_real=$(realpath -m "${d}" 2>/dev/null) || continue
+        if [[ "${d_real}" == "${toplevel_real}" ]]; then
```
kojiromike: Recommend single equals. Just a style thing here, but double equals has subtle breakage when dealing with some shells. (Only time you need == in bash is in arithmetic context.)

bradymiller: done

utilities/openemr-cmd/openemr-cmd
bradymiller: done

docker/openemr/flex/utilities/devtools
bradymiller: done on both fronts (ending up converting entire 'CODE QUALITY & STYLING' section to case/esac)

utilities/openemr-cmd/openemr-cmd
```diff
@@ -279,6 +314,8 @@ volumes:
     name: openemr-${slug}_nodemodules
   vendordir:
     name: openemr-${slug}_vendor
+  phpstanvolume:
```
kojiromike, line 317: Is `phpstanvolume` actually mounted anywhere? I don't see it referenced by any service in the base `docker-compose*.yml` files (or in this override). A top-level volume declaration without a corresponding service mount is dead — Compose won't create the volume and nothing can read or write to it. If a mount was intended (e.g. to cache the PHPStan result cache across container recreations), the service-level `volumes:` entry seems to be missing. Otherwise this declaration can be dropped.

utilities/openemr-cmd/openemr-cmd
```diff
@@ -815,6 +852,154 @@ cmd_worktree() {
...
+        echo "  Backed up existing hook to ${backup}"
+    fi
+
+    case "${stage}" in
+        commit-msg)
+            cat > "${target}" <<HOOKEOF
+#!/bin/sh
+# openemr-cmd-managed commit-msg hook -- do not edit by hand
+# ${PREK_HOOK_MARKER}
+exec "${cmd_path}" prek run --hook-stage commit-msg --commit-msg-filename "\$1"
```
kojiromike, line 912: Worth confirming this works for linked worktrees specifically. Git passes `$1` as a path relative to the worktree (commonly `.git/COMMIT_EDITMSG`), and that path then gets handed unmodified to `pre-commit` inside the container, where it resolves against the in-container cwd (`/var/www/localhost/htdocs/openemr`). The earlier override comments note the primary's `.git` is bind-mounted at the same absolute host path, but linked worktrees use a gitfile pointing to `<gitcommondir>/worktrees/<name>/` — has that path been validated as resolvable from inside the container, both when git passes a relative path and when it passes an absolute one? Failure mode would be a silent `pre-commit: file not found` rather than anything obvious.

kojiromike (commented):
Nice cleanup — the if-cascade → case/function dispatch in devtools reads much better, the apk install hoisting genuinely fixes the down --keep-volumes && up regression, and the prek dispatcher is a real ergonomic win for host-side commits. Two questions inline plus one operational nit.

Operational: the test plan is comprehensive but all items are unchecked. Could you confirm which scenarios you've actually exercised end-to-end? In particular the actionlint-docker → actionlint-system substitution during a real git commit from a worktree, and the package-persistence check after down --keep-volumes && up.

utilities/openemr-cmd/openemr-cmd
bradymiller: hi @kojiromike , Noted this dir created by phpstan will break things on worktrees etc and in related document PR in the openemr repo: https://github.com/openemr/openemr/pull/12147

bradymiller:
testing well and bringing in

Issue #758, opened 2026-05-30
fix(release): release App needs `workflows:write` on openemr-devops (rotation push to build-*.yml rejected)


kojiromike:
## Problem

`release-rotation.yml` force-pushes `release-rotation/auto`, and that branch rewrites `.github/workflows/build-{800,810,811}.yml` — these are registered as `kind: build_workflow` in `tools/release/versions.yml`, so rotation updates the docker-tag pins inside them. The release GitHub App's installation on `openemr/openemr-devops` has `contents:write` + `pull-requests:write` but **not** `workflows:write`, so the push is rejected:

```
! [remote rejected] release-rotation/auto -> release-rotation/auto
  (refusing to allow a GitHub App to create or update workflow
   .github/workflows/build-800.yml without `workflows` permission)
```

Because the push fails, the `release-rotation/auto` branch and PR are never created — the infra slice of the three-PR release flow is missing, and `ship-release.yml` has no infra PR to merge. Read-only consumers on the same dispatch (`release-announcements.yml`) succeed, which is why this only surfaces on the rotation push.

## Reproduction

- `openemr-tag` dispatch run `26594165691` (2026-05-28), step "Force-push rotation branch" → exit 201.
- Same failure on `26535911455` (2026-05-27).

## Fix (admin)

- [x] Grant the release GitHub App `workflows:write` (Repository permissions → Workflows: Read & write) and approve the new permission on the `openemr/openemr-devops` installation.

## Hardening (follow-up PR)

- [ ] `docs/release-automation-plan.md`'s permissions self-check lists only `contents:write` + `pull-requests:write`; add `workflows:write` with the rotation-registry justification.
- [ ] `release-permissions-check.yml` writes a plain dotfile, so it can't catch workflow-file push perms; extend it to PUT a stub `.github/workflows/*.yml` so this drift surfaces in the self-check.

## Refs

openemr/openemr-devops#664, #706, #701; `docs/RELEASE_PROCESS.md` (Bot identity and credentials).

bradymiller:
hi @kojiromike , just did this

PR #759, opened 2026-05-30
refactor(release): port App permission probes from shell to PHP CLI

kojiromike:
## Problem

`release-rotation.yml` rewrites `.github/workflows/build-{800,810,811}.yml` (registered as `kind: build_workflow` in `tools/release/versions.yml`), so rotation legitimately updates workflow files. The release App was missing `workflows:write`, so the rotation push was rejected at runtime:

> refusing to allow a GitHub App to create or update workflow `.github/workflows/build-800.yml` without `workflows` permission

The permission self-check only created plain dotfiles on a throwaway branch, so it exercised `contents:write` but never `workflows:write` — the gap went undetected until rotation failed (#758).

## Change

The five `tools/release/scripts/probe-app-*.sh` probes were the only untested, non-PHP logic in an otherwise all-PHP, phpstan(level 10) / phpcs / phpunit-gated tool. Rather than pile more shell on, this ports them into the existing tested architecture:

- `AppPermissionApi` / `GhAppPermissionApi` — the narrow GitHub surface (gh + Symfony Process) the probes touch.
- `AppPermissionProbe` — maps a low-level failure to the specific missing permission, including the `workflows:write` gap: a commit under `.github/workflows/` on the throwaway branch is refused without it, which a plain-dotfile probe misses.
- `bin/probe-app.php` — `SingleCommandApplication` dispatching by `--check`, defaulting owner/repo-name/branch/run-id from the workflow env.
- `FakeAppPermissionApi` + `AppPermissionProbeTest` cover each permission passing and each failure path (including the workflow-file refusal).

`Taskfile.yml` `release:probe-app-*` targets now run the PHP binary (with the `setup` dep), and `release-permissions-check.yml` gains a Setup PHP step. The shell scripts are removed.

## Note

The `contents:write` / `workflows:write` probe step fails until the App is granted `workflows:write` (#758); that failure is intentional — it surfaces the gap rather than letting it resurface as a rotation push rejection.


tools/release/scripts/probe-app-contents-write.sh
```diff
@@ -23,6 +27,18 @@ main() {
...
     fi
+    # Inert workflow stub: name-only, no `on:` trigger, so it never schedules
+    # or runs. Push is rejected by GitHub if the App lacks workflows:write,
+    # which is the exact failure mode that broke release-rotation.yml when
+    # rewriting build-800.yml.
+    workflow_b64=$(printf 'name: permissions-check %s\n' "${RUN_ID}" | base64)
+    if ! gh api -X PUT "/repos/${OWNER}/${REPO_NAME}/contents/.github/workflows/permissions-check-${RUN_ID}.yml" \
+            -f "message=permissions-check workflow ${RUN_ID}" \
+            -f "content=${workflow_b64}" \
+            -f "branch=${BRANCH}" >/dev/null; then
```
kojiromike: Does this need permission to push to `$BRANCH`, or just any branch? Should we use a throwaway branch?

kojiromike: Resolved by the PHP port (68587cf): this shell line is gone. For the record — `workflows:write` is not branch-scoped (GitHub enforces it on any create/update under `.github/workflows/` on any branch), and `$BRANCH` was already a per-run throwaway (`release-permissions-check/<run-id>`, created at start and deleted by the cleanup probe). `AppPermissionProbe::checkContentsWrite` keeps that same throwaway-branch behavior.

PR #762, opened 2026-05-31
fix(openemr-cmd): keep worktree list iterating past entries with deleted dirs

bradymiller:
## Summary
- `openemr-cmd worktree list` was silently truncating its output, stopping at the first state entry whose directory had been deleted out-of-band.
- Root cause: under `set -euo pipefail`, the pattern `wt_validate_dir_safe "${dir}"; validate_result=$?` is two separate commands, so `set -e` exits on the non-zero return before `validate_result=$?` runs.
- Switched to `wt_validate_dir_safe "${dir}" || validate_result=$?` — the failing call now sits in an `||` list, which `set -e` ignores, so the loop completes and every entry prints (bad ones as `missing` / `invalid`).

## Test plan
- [x] Reproduced against a `.worktrees.json` with 6 entries where the 5th entry's dir had been removed — pre-fix output stopped at 4 entries with exit code 1; post-fix prints all 6 with exit code 0, missing dir correctly labeled `missing`.
- [x] `.worktrees.json` not mutated by the list command.

Assisted-by: Claude Code

### DIFF
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -18,7 +18,7 @@ set -euo pipefail
 #################################################################################################
 
 # Increment the version when modify script
-VERSION="1.0.34"
+VERSION="1.0.35"
 
 # ============================================================================
 # WORKTREE STATE MANAGEMENT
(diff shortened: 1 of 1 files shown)


bradymiller:
testing well and bringing in

Issue #638, opened 2026-04-08
Release process upgrades


Firehed:
Following 2026-04-08 release call, we have some ideas for improving the release process

- Aim for a `next` (whatever upcoming version is in flight, the RC basically), `dev` (head of master), and a `current` (the most recent tagged release) in the devops CI tooling
- consolidate some of the tests around these naming changes; aim to make them more reusable where possible

Aim to release 1 week after the branch. Security patches in between. Version numbers as examples given `8.1.0` as the latest OpenEMR release.

```
Week 0: Apply security patches to release branch and master, tag `8.1.1`. Make `rel-8.2` from `master`. Adjust infra to reflect new `next` and `current`.
During week 0: Test on `rel-8.2`, commit fixes and ensure they end up on that branch and `master`.
Week 1: Assuming no standing showstoppers, tag `8.2.0` from the head of `rel-8.2`. Adjust infra as needed.
Week 2: Apply security patches and tag `8.2.1`
Week 3: Apply security patches and tag `8.2.2`
Week 4: Apply security patches and tag `8.2.3`, branch `rel-8.3` off master (week 0 procedure)
```

RC branches should eventually trigger all of the build automation.

Official docker distribution should mostly stick with latest version of OS+PHP. Builds on older versions are more for compatibility testing but not really for distribution.

Look into Flysystem (https://flysystem.thephpleague.com/docs/) as way to solve document storage abstraction (sort of what CouchDB aims/aimed to do). This enables getting document storage out of webroot, which itself makes upgrades WAY easier and safer.

PR #777, opened 2026-06-03
ci(build): consolidate rotating build workflows + publish version.php-derived tags

kojiromike:
## Summary

Collapse the three version-named rotating build workflows into one slot-named matrix workflow, then make that workflow publish each slot's **true** OpenEMR version (derived from `version.php`) instead of the bare symlink directory name.

### 1. Consolidate to a single slot-matrix workflow

`build-800.yml` / `build-810.yml` / `build-811.yml` become a single `build-openemr.yml` whose matrix is the **stable slot names** `[current, next, dev]` × platform — never rotated. Each job resolves the slot's version at runtime from the `docker/openemr/{current,next,dev}` symlinks, so the workflow holds zero version strings and rotation never rewrites it (`SlotRotator` just re-points the symlink).

Docker dirs (`docker/openemr/{8.0.0,8.1.0,8.1.1}`) stay version-named — they hold real per-version Dockerfiles and are honest targets.

### 2. Publish version.php-derived tags per slot

The merge job derives each slot's published version from the OpenEMR `version.php` at the slot Dockerfile's `OPENEMR_VERSION` ref, composing `major.minor.patch[.realpatch][tag]`:

| Slot | ref | Published `:ver` | Was (bare dir) |
|---|---|---|---|
| current | `rel-800` | **8.0.0.3** | 8.0.0 |
| next | `rel-810` | 8.1.0 | 8.1.0 |
| dev | `master` | **8.1.1-dev** | 8.1.1 |

This fixes two bugs: `current` is really 8.0.0.3 (a genuine patch release), and `dev` would otherwise publish a bare `:8.1.1` that collides with the eventual real 8.1.1 release.

**Tag set per slot:** `:<float>` (latest/next/dev) + `:<ver>` + `:<ver>-<date>`, plus the bare-dir tag (`8.0.0` / `8.1.0`) for **current/next only** (never dev) when it differs from `:<ver>`.

### Supporting changes

- Standardize `docker/openemr/8.0.0/Dockerfile` to `ARG OPENEMR_VERSION=rel-800` so the merge-job grep is uniform across all three slots.
- Registry `tools/release/versions.yml`: `current.full` → `8.0.0.3` (drop vestigial `patch`), `dev.full` → `8.1.1-dev`, 8.0.0 Dockerfile `kinds` → `docker_arg_branch`.
- Docker Hub overview template + renderer test updated to the `docker_dir` + `full` model.
- `docs/release-automation-plan.md` documents the version.php derivation and per-slot tag set.

This is the build-workflow half of step #5 (workflow consolidation, #638 follow-on) in `docs/release-automation-plan.md`, and unblocks the release-rotation PR (#760).

## Test plan

- [x] `composer test` — 179 tests pass
- [x] `composer phpcs` — clean
- [x] `composer phpstan` — no errors
- [x] `actionlint` on `build-openemr.yml` — clean (shellcheck SC2016 suppressed for the PHP one-liner, matching `build-release.yml`)
- [x] `yq` lint on the registry
- [x] Version-composition dry-run against all three refs → 8.0.0.3 / 8.1.0 / 8.1.1-dev
- [ ] CI green on this PR


.github/workflows/build-openemr.yml
```diff
@@ -0,0 +1,146 @@
...
+            bare_dir: false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Resolve slot version
+        id: resolve
+        run: |
+          dir="$(basename "$(readlink "docker/openemr/${{ matrix.slot }}")")"
+          ref="$(grep -oE 'OPENEMR_VERSION=[A-Za-z0-9._-]+' "docker/openemr/${dir}/Dockerfile" | head -1 | cut -d= -f2)"
```
kojiromike: The `head -1` grep isn't anchored to the `ARG` directive, so it can match a *comment* that mentions `OPENEMR_VERSION=`. `docker/openemr/8.1.0/Dockerfile` has `#   - Branch: --build-arg OPENEMR_VERSION=master` (line 154) before the real `ARG OPENEMR_VERSION=rel-810` (line 156), so the `next` slot resolves `ref=master`. The merge job then fetches `version.php` from master and derives `VER=8.1.1-dev` — mis-versioning the 8.1.0 image and publishing `:8.1.1-dev` / `:8.1.1-dev-<date>` tags that collide with the `dev` slot. (`dev` happens to work only because its real ARG is also `master`.)

Reproduced locally: current=rel-800, **next=master**, dev=master. Anchoring to the directive yields the correct rel-800 / rel-810 / master:

```suggestion
          ref="$(grep -oE '^ARG OPENEMR_VERSION=[A-Za-z0-9._-]+' "docker/openemr/${dir}/Dockerfile" | head -1 | cut -d= -f2)"
```

docker/openemr/dev
```diff
@@ -0,0 +1 @@
+8.1.1
```
kojiromike, line 1: These new slot symlinks (`current`/`next`/`dev`) get swept into the production-test matrix. `test-core.yml` discovers targets with an unrestricted glob — `dirs=( docker/openemr/*/Dockerfile )` (line 40) — so each symlink resolves to a dir-with-Dockerfile and becomes an extra `Production Docker (<slot>)` job that rebuilds 8.0.0/8.1.0/8.1.1 redundantly. `dev` (→ 8.1.1 → master) then trips the upstream `RootCliGuard` ("CLI scripts must not be run as root") and fails CI on this PR. The same guard already fails `flex`/`8.1.1`, so it isn't this PR's root cause — but the `dev` alias adds another red X plus wasted CI on the `current`/`next` duplicates.

Suggest restricting discovery to skip symlinked dirs in `test-core.yml` (not in this diff, so a small scope add):

```bash
dirs=()
for f in docker/openemr/*/Dockerfile; do
  d=${f%/Dockerfile}
  [ -L "$d" ] || dirs+=( "$d" )
done
```

.github/workflows/build-openemr.yml
kojiromike: Fixed in cf81a6b — anchored the grep to `^ARG OPENEMR_VERSION=` so the comment line no longer matches. Verified locally: current=rel-800, next=rel-810, dev=master.

docker/openemr/dev
kojiromike: Fixed in 800ece4 — `test-core.yml` discovery now skips symlinked dirs. Confirmed on the latest run: the Production Docker matrix is `7.0.4, 8.0.0, 8.1.0, 8.1.1, binary, flex` with no `current`/`next`/`dev` jobs.

docs/release-automation-plan.md
```diff
@@ -85,10 +85,10 @@ host most of the rewrite logic.
    this repo, accepted via `repository_dispatch` from openemr/openemr.
 
 5. **Workflow consolidation (#638 follow-on).**
-   - Collapse `build-810.yml` / `build-811.yml` / `build-800.yml` etc. into one
-     `build-current.yml` / `build-next.yml` / `build-dev.yml` driven by the
-     registry. Reduces what the rotation script has to touch.
-   - Same for `test-flex-*.yml` matrices.
+   - Done: the rotating build workflows are slot-named `build-current.yml` /
+     `build-next.yml` / `build-dev.yml`, driven by the registry, so rotation
+     rewrites the pins inside without leaving misleadingly version-named files.
```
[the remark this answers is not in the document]
kojiromike: Addressed by this PR's permissions-self-check rewrite — `docs/release-automation-plan.md` no longer references `build-{800,810,811}.yml`; it now describes the consolidated `build-openemr.yml` slot model. No stale `build-8xx` references remain.

adunsulag (approved):

adunsulag:
Looks like a good automation improvement and consolidation.

PR #783, opened 2026-06-05
fix(openemr-cmd): use portable realpath for macOS/BSD

kojiromike:
## Summary

The `openemr-cmd worktree` subcommands fail on macOS with `realpath: illegal option -- e`. They pass GNU coreutils flags (`-e`, `-m`) to `realpath`, but the BSD `realpath` shipped on macOS doesn't accept them. The git worktree gets half-created on disk, but state registration (compose override + `.env` + `.worktrees.json` entry) never completes.

## Background: GNU vs BSD realpath

GNU `realpath` has three modes:
- **default** — all path components except the leaf must exist
- **`-e`** — all components *including* the leaf must exist, else fail
- **`-m`** — nothing need exist

BSD `realpath` (macOS) has a single mode equivalent to GNU `-e`: it resolves symlinks and errors on a missing leaf. Every `-m` call site in this script resolves a path that already exists (`mkdir -p` runs first, or the path was just reported by git), so bare BSD `realpath` is behaviorally equivalent at all 17 sites.

## Fix

Detect the platform once via `realpath --version` (GNU prints a version; BSD errors out) and build command arrays carrying the correct flags:

```bash
if realpath --version >/dev/null 2>&1; then
    OE_REALPATH_E=(realpath -e)
    OE_REALPATH_M=(realpath -m)
else
    OE_REALPATH_E=(realpath)
    OE_REALPATH_M=(realpath)
fi
```

Using command arrays rather than a wrapper function keeps `realpath` an external command, which avoids shellcheck `SC2310` (a function invoked in an `||` condition disables `set -e`) under the repo's `enable=all` config. Symlink resolution — the security property feeding the path-containment checks (e.g. blocking `docker/library -> /etc` escapes) — is preserved natively on both platforms.

Zero behavior change on GNU/Linux: the flags pass straight through. No new runtime dependency.

## Test plan

- [x] `shellcheck` clean under the repo's `.shellcheckrc` (`enable=all`, `--check-sourced --external-sources`)
- [x] `bash -n` passes
- [x] On macOS (BSD realpath, bash 3.2): `openemr-cmd worktree add <branch> -b` completes without `realpath: illegal option` — compose override + `.env` written, `.worktrees.json` entry registered
- [x] Verified array idiom works on macOS bash 3.2 (arrays always non-empty under `set -u`)

### DIFF
--- a/tests/bats/openemr-cmd/helpers.bash
+++ b/tests/bats/openemr-cmd/helpers.bash
@@ -62,7 +62,7 @@ STUB
 # If the script is restructured, bump this constant — the test
 # 'OC_SCRIPT_FUNCS_END points at end of function defs' in helpers_pure.bats
 # will fail loudly when it drifts.
-OC_SCRIPT_FUNCS_END=1446
+OC_SCRIPT_FUNCS_END=1460
 
 # Source ONLY the function definitions of openemr-cmd into the current shell.
 # Caller is responsible for setting OPENEMR_ROOT (and WT_STATE_FILE / others
--- a/utilities/openemr-cmd/openemr-cmd
+++ b/utilities/openemr-cmd/openemr-cmd
@@ -18,7 +18,7 @@ set -euo pipefail
 #################################################################################################
 
 # Increment the version when modify script
-VERSION="1.0.38"
+VERSION="1.0.39"
 
 # ============================================================================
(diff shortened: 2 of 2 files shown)


bradymiller:
nice!

PR #785, opened 2026-06-06
ci(openemr-cmd): add macOS leg to BATS matrix; fix bash 3.2 compat

kojiromike:
## Summary

`openemr-cmd` is a host CLI developers run on their Macs, but its BATS workflow only ran on `ubuntu-22.04`. So macOS-portability bugs reached users with zero CI coverage — most recently the BSD `realpath` failure fixed manually in #783. This adds a `macos-14` leg to the BATS matrix, where the script runs under macOS system **bash 3.2**, the sharpest portability risk.

Adding that leg immediately surfaced two real bash 3.2 incompatibilities, both fixed here:

- **`openemr-cmd`** used `${RESP,,}` (bash 4+ case-modification), which throws `bad substitution` under bash 3.2. Replaced with a `[Yy]` glob match in `[[ ]]`, preserving the `[Y/n]` default-yes semantics. (Verified identical behavior on bash 3.2 and 5.)
- **Test harness** sourced the script via process substitution (`source <(head -n …)`), which silently defines no functions under macOS bash 3.2. Switched the three sites (`helpers.bash`, `helpers_pure.bats`, `state.bats`) to `eval "$(head -n …)"`.

The repo is public, so GitHub-hosted macOS minutes are free. The macOS leg is **advisory** for now (`fail-fast: false`, not branch-protection-required) and can be promoted to required later.

## Test plan

- [x] Full openemr-cmd BATS suite passes **39/39 under system bash 3.2** (macOS-runner equivalent)
- [x] Full suite passes **39/39 under bash 5** (Linux-runner equivalent)
- [x] `${RESP,,}` → `[Yy]` glob verified semantically identical across both shells (empty→pull, y/Y→pull, else→skip)
- [x] Workflow YAML validates; `helpers.bash` and `openemr-cmd` shellcheck-clean
- [ ] CI: confirm both `ubuntu-22.04` and `macos-14` legs go green

### DIFF
--- a/.github/workflows/test-bats-openemr-cmd.yml
+++ b/.github/workflows/test-bats-openemr-cmd.yml
@@ -16,8 +16,12 @@ on:
 
 jobs:
   bats-openemr-cmd:
-    name: BATS openemr-cmd
-    runs-on: ubuntu-22.04
+    name: BATS openemr-cmd (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macos-14]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v6
       - name: Install Bats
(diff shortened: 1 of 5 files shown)


bradymiller:
nice!

PR #815, opened 2026-06-23
feat(openemr-cmd): add --base for worktree add, default to canonical openemr/openemr master

### LINKED ISSUE #814: openemr-cmd worktree add: support --base <ref> to branch off non-master refs
## Problem

`openemr-cmd worktree add <new-branch> -b` always bases the new branch off the **primary repo's current HEAD**, which is master in practice. The relevant code in `openemr-cmd`:

```bash
git -C "${OPENEMR_ROOT}" worktree add -b "${branch}" "${dir}"
```

`git -C "${OPENEMR_ROOT}"` runs in the primary repo root, and `git worktree add -b <branch> <path>` with no `<commit-ish>` uses that repo's HEAD. There's no way to ask for a different base ref.

This is fine for normal feature work off master, but bites when working on a feature targeted at a **rel branch** (e.g., preparing a per-release-on-rel-branch fix). To branch off `rel-810`, today the options are all unpleasant:

1. **In-place inside the rel-810 worktree:** `openemr-cmd worktree add rel-810` (no `-b`), then inside it `git checkout -b <feature>` + reset the worktree's `rel-810` back to upstream. Works but the worktree DIR is named `openemr-wt-rel-810` (after the parent rel branch, not the feature), and rel-810 transiently sits as the worktree's checked-out branch — easy to accidentally commit onto it.
2. **Park primary on rel-810:** `cd <primary> && git checkout rel-810 && git pull upstream rel-810 && openemr-cmd worktree add <feature> -b && git checkout master`. Works for single-user but has the same multi-agent coordination problem already documented for the `--update-head-ok` case — a concurrent agent's `openemr-cmd worktree add -b` would silently base off rel-810 instead of master.
3. **Raw `git worktree add -b <feature> <path> rel-810`:** explicitly forbidden by CLAUDE.md (orphans openemr-cmd state — no `.worktrees.json` entry, no compose override, no port offset).

None of these are good.

## Proposed solution

Extend the `add` subcommand with a `--base <ref>` flag:

```
openemr-cmd worktree add <new-branch> -b [--base <ref>]
```

- Default: `origin/master` (or `upstream/master` — whichever is the configured upstream-tracking convention; today this would change current behaviour from \"primary's HEAD\" to \"a freshly-fetched master\", which is actually a strict improvement per the planning notes below)
- `--base rel-810` → fetches the ref, then runs `git -C \${OPENEMR_ROOT} worktree add -b <new-branch> <dir> <fetched-ref>`
- `--base upstream/rel-800`, `--base origin/rel-704`, `--base v8_0_0_3`, `--base <sha>` all work because git's `<commit-ish>` accepts any of them

Internally:

```bash
if [[ -n \"\${base}\" ]]; then
    # Optionally: fetch first if base looks like a remote ref
    git -C \"\${OPENEMR_ROOT}\" fetch \"\${base_remote}\" \"\${base_ref}\"   # if applicable
    git -C \"\${OPENEMR_ROOT}\" worktree add -b \"\${branch}\" \"\${dir}\" \"\${resolved_base}\"
else
    # current behaviour (or: default to fetching + basing off origin/master)
    git -C \"\${OPENEMR_ROOT}\" worktree add -b \"\${branch}\" \"\${dir}\"
fi
```

## Related work

There's an existing planned tool-level fix to make the default-`-b` flow auto-fetch + base off `origin/master`, motivated by the same multi-agent concern that drove the CLAUDE.md `--update-head-ok` ban. That fix and `--base` are the same internal change — both replace the implicit \"primary's HEAD\" base with an explicit, fetched ref. Worth implementing them together:

- Default (no `--base`): fetch `origin/master`, base off it
- `--base <ref>`: fetch the named ref (if remote), base off it
- Either way: primary repo's HEAD is never read or modified, so it stays a non-coordinating surface for multi-agent setups

## Use case driving this

Preparing per-release work on a rel branch — bumping docker upgrade machinery, version.php, etc. Recent example: 8.1.1 release prep on rel-810 (PRs openemr/openemr#12608 + #12609) hit option 1 above; the workflow was viable but the rel-810 branch's transient presence in the worktree's checked-out-branch slot was a sharp edge that motivated this request.

## Acceptance

- [ ] `openemr-cmd worktree add <feature> -b --base rel-810` creates a worktree at `openemr-wt-<feature>` checked out on `<feature>` based on rel-810's tip
- [ ] Primary repo's HEAD is never modified during the operation
- [ ] Default (no `--base`) fetches `origin/master` and bases off it (current behaviour preserved in spirit but no longer reads primary HEAD)
- [ ] Help text + `CONTRIBUTING.md`/`README.md` updated

bradymiller:
Resolves #814.

## Summary

Before this change, `openemr-cmd worktree add <branch> -b` based the new branch on the primary repo's HEAD. That made primary HEAD an implicit coordination surface across concurrent agents and silently stale when the primary hadn't been freshly fetched.

This PR replaces that with an explicit, freshly-fetched canonical default and a `--base <ref>` flag with **two-form semantics**:

- **`--base <url>[#<ref>]`** — fetch from URL. For AI assistants and anyone who wants a fully-reproducible, network-fetched base independent of local state. Default ref is `master`. `file://` is accepted (which is how the bats tests work without network).
- **`--base <commit-ish>`** — pure git `rev-parse` semantics. Handles local branches, remote-tracking refs (`origin/master`), tags, SHAs, `HEAD`, `@{u}`, anything git understands. No fetch — freshness is the caller's responsibility (just like `git checkout -b ... <start-point>`).

Default behavior (no `--base`) still fetches canonical `openemr/openemr` master directly, so the ergonomic "just give me a worktree" path is unchanged. Primary HEAD is never read on the default path.

## Why two forms

[Discussion in the issue thread](https://github.com/openemr/openemr-devops/issues/814) converged on two distinct usage profiles:

| Profile | What they pass | What they get |
|---|---|---|
| Low-scale human dev | `--base <bare-ref>` | Pure git semantics — local branch / tag / SHA. Fast, no network. |
| High-scale AI agent | `--base https://github.com/openemr/openemr.git#<ref>` | Explicit canonical fetch every time. Network, reproducible across agents/machines, zero coupling to local state. |

This is cleaner than the earlier "fetch on bare names" sketch (which forced a `local:` escape hatch for local refs) and aligns `--base` with how git's native `<commit-ish>` works.

## Behavior matrix

| Invocation | What happens |
|---|---|
| `... -b` | Fetch `master` from canonical `https://github.com/openemr/openemr.git`; base on resolved SHA |
| `... -b --base https://github.com/openemr/openemr.git#rel-810` | Fetch `rel-810` from that URL; base on resolved SHA |
| `... -b --base file:///path/to/local-fork#branch` | Fetch from `file://` URL; base on resolved SHA |
| `... -b --base master` | Local `master`, no fetch |
| `... -b --base origin/master` | Local remote-tracking ref `origin/master`, no fetch |
| `... -b --base v8_0_0_3` | Local tag, no fetch |
| `... -b --base HEAD` | Whatever primary `HEAD` is (explicit opt-in) |
| `... -b --base <sha>` | That SHA, no fetch |
| `... -b --base nonexistent-xyz` | Fail with hint: re-run as `--base https://github.com/openemr/openemr.git#nonexistent-xyz` to fetch it |
| `... --base master` (no `-b`) | Fail: `--base only applies with -b` |

`--no-track` is passed to `git worktree add` so the new branch doesn't inherit the base ref's remote-tracking config — matches the previous no-tracking behavior.

`WT_CANONICAL_URL` env var overrides the canonical URL for forks/mirrors.

## Concurrency safety

Fetches go through `wt_fetch_to_sha`, which writes into a per-invocation temp ref (`refs/wt-base-tmp/<pid>-<rand>-<rand>`) and immediately resolves to an immutable SHA before any `git worktree add` consumes it. This avoids the `FETCH_HEAD` race that would otherwise let two concurrent agents overwrite each other's fetch result. Temp refs are cleaned up best-effort; orphans under `refs/wt-base-tmp/*` are harmless.

## Files

- `utilities/openemr-cmd/openemr-cmd` — `wt_resolve_base()` + `wt_fetch_to_sha()` helpers, `--base` parsing, `WT_CANONICAL_URL` constant, version → `1.0.42`, updated inline help on both worktree help blocks (top-level `-h` + `cmd_worktree()` no-subcommand fallback)
- `utilities/openemr-cmd/openemr-cmd-h` — `-A` line count bump for new help block
- `tests/bats/openemr-cmd/worktree_add_base.bats` — 16 bats cases covering both `--base` forms, all error paths, and a drift guard that asserts both help blocks surface `[--base <ref>]`
- `tests/bats/openemr-cmd/helpers.bash` — `OC_SCRIPT_FUNCS_END` bumped for the new helpers

## Paired follow-up — landed

openemr/openemr#12613 — paired doc update to `openemr/CONTRIBUTING.md` and `openemr/CLAUDE.md` explaining the new `-b` default and `--base` flag, so contributors and AI agents reading those docs don't assume the old HEAD-based behavior.

## Test plan

- [x] `bats tests/bats/openemr-cmd/` — 60 passing (16 new, 44 existing, 0 failures)
- [x] CI: BATS (ubuntu-22.04 + macos-14), ShellCheck — all green
- [x] On a real install: `openemr-cmd worktree add new-feature -b` creates a worktree based on freshly-fetched canonical master
- [x] `--base https://github.com/openemr/openemr.git#rel-810` bases on canonical rel-810
- [x] `--base master` bases on local master (no network)
- [x] `--base upstream/master` bases on local remote-tracking ref
- [x] `--base nonexistent-xyz` errors with the canonical-URL hint
- [x] `--base master` without `-b` errors clearly
- [x] Primary repo's HEAD is unchanged before/after in all cases above
- [x] `openemr-cmd -h` shows the new options (openemr-cmd-h unchanged structurally; just bumped -A line count)
- [x] `openemr-cmd worktree` (no subcommand) fallback also shows the new options
- [x] Two concurrent `openemr-cmd worktree add ... -b` calls each get their own base SHA (no FETCH_HEAD race) — verified with parallel `--base <url>#rel-810` + `--base <url>#rel-800`, each landed correctly

### DIFF
--- a/tests/bats/openemr-cmd/helpers.bash
+++ b/tests/bats/openemr-cmd/helpers.bash
@@ -62,7 +62,7 @@ STUB
 # If the script is restructured, bump this constant — the test
 # 'OC_SCRIPT_FUNCS_END points at end of function defs' in helpers_pure.bats
 # will fail loudly when it drifts.
-OC_SCRIPT_FUNCS_END=1463
+OC_SCRIPT_FUNCS_END=1536
 
 # Source ONLY the function definitions of openemr-cmd into the current shell.
 # Caller is responsible for setting OPENEMR_ROOT (and WT_STATE_FILE / others
--- a/tests/bats/openemr-cmd/helpers.bash
+++ b/tests/bats/openemr-cmd/helpers.bash
@@ -62,7 +62,7 @@ STUB
 # If the script is restructured, bump this constant — the test
 # 'OC_SCRIPT_FUNCS_END points at end of function defs' in helpers_pure.bats
 # will fail loudly when it drifts.
-OC_SCRIPT_FUNCS_END=1536
+OC_SCRIPT_FUNCS_END=1563
 
 # Source ONLY the function definitions of openemr-cmd into the current shell.
 # Caller is responsible for setting OPENEMR_ROOT (and WT_STATE_FILE / others
--- a/tests/bats/openemr-cmd/worktree_add_base.bats
+++ b/tests/bats/openemr-cmd/worktree_add_base.bats
@@ -1,0 +1,174 @@
+# BATS: --base flag of `openemr-cmd worktree add` and the underlying
+# wt_resolve_base helper.
+#
+# Contract under test:
+#   --base <url>[#<ref>]    -> fetch from URL, base on FETCH_HEAD
+#   --base <git commit-ish> -> pure git rev-parse resolution, no fetch
+#   no --base               -> fetch canonical openemr/openemr master
+#                              (NOT routed through wt_resolve_base; covered
+#                              via the integration-error tests only)
+
+load '../test_helper/bats-support/load'
+load '../test_helper/bats-assert/load'
+load 'helpers'
+
+setup() {
+    SCRIPT="$(oc_script_path)"
+    [[ -x "$SCRIPT" ]] || skip "openemr-cmd script not found"
+    TMP_ROOT=$(oc_mktempdir)
+
+    # Real local git repo with one commit on master, an extra branch, and
+    # a tag. Used both as OPENEMR_ROOT (target of rev-parse) and as a
+    # file:// fetch source for URL-form tests.
+    oc_init_repo "${TMP_ROOT}"
+    git -C "${TMP_ROOT}" branch rel-test
+    # tag.gpgsign / tag.forceSignAnnotated in a user's global config would
+    # otherwise force annotated+signed tags and prompt for a message.
+    git -C "${TMP_ROOT}" -c tag.gpgsign=false -c tag.forceSignAnnotated=false tag v-test
+
+    export TMP_ROOT
+}
+
+teardown() {
+    [[ -n "${TMP_ROOT:-}" ]] && rm -rf "${TMP_ROOT}"
+    [[ -n "${STUB_DIR:-}" ]] && rm -rf "${STUB_DIR}"
+    # Some tests don't set STUB_DIR, so the [[ -n ]] above can be the
+    # teardown's last command and return 1. Force success.
+    return 0
+}
+
+# Run a wt_resolve_base call in a subshell with the script's function defs
+# sourced. Stderr is folded into stdout so assertions on hint messages work.
+run_resolve() {
+    local arg="$1"
+    run env \
+        OPENEMR_ROOT="${TMP_ROOT}" \
+        WT_CANONICAL_URL="https://github.com/openemr/openemr.git" \
+        bash -c "
+            set -uo pipefail
+            eval \"\$(head -n ${OC_SCRIPT_FUNCS_END} '${SCRIPT}')\"
+            wt_resolve_base '${arg}' 2>&1
+        "
+}
+
+# --- URL form ---------------------------------------------------------------
+
+@test "wt_resolve_base: file:// URL with #ref fetches and echoes FETCH_HEAD" {
+    run_resolve "file://${TMP_ROOT}#master"
+    assert_success
+    # FETCH_HEAD appears on the last line; the wt_info chatter is above it.
+    [[ "${lines[-1]}" = "FETCH_HEAD" ]]
+}
+
+@test "wt_resolve_base: file:// URL without #ref defaults to master" {
+    run_resolve "file://${TMP_ROOT}"
+    assert_success
+    [[ "${lines[-1]}" = "FETCH_HEAD" ]]
+}
+
+@test "wt_resolve_base: URL fetch failure returns non-zero with 'fetch failed' hint" {
+    run_resolve "file:///nonexistent/repo/that/does/not/exist.git#master"
+    assert_failure
+    assert_output --partial "fetch failed"
+}
+
+@test "wt_resolve_base: git@ form is dispatched as URL (not as a local ref)" {
+    # We can't actually auth in the test env, so we expect the fetch step
+    # to fail. The point is to confirm git@... gets to the fetch branch
+    # rather than falling through to rev-parse with the 'is not a URL' hint.
+    run_resolve "git@github.com:fake-org-bats/fake-repo-bats.git#master"
+    assert_failure
+    refute_output --partial "is not a URL"
+}
+
+@test "wt_resolve_base: https:// form is dispatched as URL" {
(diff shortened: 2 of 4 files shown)


bradymiller:
Revised per discussion on #814. Key change in this commit:

`--base` now has **two clean forms** instead of the previous four-way dispatch:

1. **`--base <url>[#<ref>]`** — fetch from URL. The form AI assistants are expected to use for reproducibility.
2. **`--base <commit-ish>`** — pure git `rev-parse` semantics. Local branches, remote-tracking refs, tags, SHAs, `HEAD`, `@{u}` — anything git understands. No fetch.

This eliminates the "bare name = fetch from canonical" interception that would have required a `local:` prefix as an awkward escape hatch for local refs. Two profiles, two forms, no escape hatch needed.

Default (no `--base`) still fetches canonical `openemr/openemr` master, so the ergonomic common case is unchanged and primary HEAD is still never read.

Also adds `tests/bats/openemr-cmd/worktree_add_base.bats` with 14 cases covering both `--base` forms and the error paths. Full suite is green: 58 ok, 0 failures.

bradymiller:
Pushed 9de4288 addressing review feedback + the second drift point you spotted.

**1. FETCH_HEAD race (CodeRabbit, major).** Confirmed real — `FETCH_HEAD` is one file in `.git`, so two concurrent agents fetching at once race on it and one's `worktree add` can pick up the other's commit. Same hazard as the original HEAD problem, just relocated. New `wt_fetch_to_sha` helper fetches into `refs/wt-base-tmp/<pid>-<rand>-<rand>`, immediately resolves to an immutable SHA, cleans up the temp ref. Applied to both the URL form and the default-canonical path.

**2. Second help-block drift.** `cmd_worktree()` no-subcommand fallback (line 1049-1063) had an old copy of the `add` usage line — fixed. Added two bats tests asserting both blocks contain `[--base <ref>]` so any future drift fails CI loudly.

**3. macOS bash 3.2 (CodeRabbit).** `${lines[-1]}` negative-index subscript replaced with `assert_line --regexp` which works on both bash versions.

**4. ShellCheck SC2310.** Silenced with explanation; the functions use explicit `return 1`, not `set -e`, so the warning's underlying concern doesn't apply.

Local validation: **60/60 bats tests pass; shellcheck clean.**

PR #823, opened 2026-06-24
fix(openemr-cmd): serialize state mutations behind mkdir-based state lock

bradymiller:
Fixes the race condition discovered in #820's `worktree_concurrent.bats` test 1 — un-skipping that test is this PR's success criterion.

## The race

`wt_state_set` and `wt_next_offset` do non-atomic read-modify-write on `.worktrees.json` (jq read → tmp write → mv). Two concurrent invocations both read the same starting state, each compute their independent mutation, then race on the final `mv`. Two failure modes:

| Mode | Symptom |
|---|---|
| **State entry lost** | Later `mv` clobbers earlier writer's file. Earlier branch has a worktree dir + compose files + git registration but no state entry. Tool can't manage it (`list`/`up`/`down`/`exec`/`remove`/`regen` all fail with "No worktree found"). |
| **Offset collision** | Both `wt_next_offset` calls see the same starting state and return the same offset. Both writes survive but share an offset → port collision when starting the stacks. |

Race window spans `wt_next_offset` → `wt_state_set` in `cmd_worktree_add` (covers canonical fetch, git worktree add, compose file writes — easily seconds to a minute). Real impact in the multi-agent workflows the recent `--base` PR series was hardening; zero impact for single-user single-agent.

## The fix

Serialize the state-mutating critical sections of `cmd_worktree_{add,remove,set_env}` behind a process-wide lock.

Lock primitive: a regular FILE created atomically via `set -C` (noclobber) + redirect:

```bash
( set -C; echo $$ > "${WT_STATE_LOCK_FILE}" )
```

`set -C` scoped to the subshell so the option doesn't leak. The redirect opens with `O_CREAT|O_EXCL` at the syscall level — atomic create-or-fail, only one of N concurrent acquirers wins. The winner's PID is in the lockfile from the moment it appears (the subshell creates and writes before returning), so any observer of the lockfile sees the holder identity.

| Property | Mechanism |
|---|---|
| Timeout | `WT_STATE_LOCK_TIMEOUT_S` (default 300s). Checked at the top of every acquire-loop iteration so an I/O failure can't trap us in an unbounded spin. Error message includes the lockfile path and `rm -f` recovery hint. |
| Crash safety | `EXIT` trap releases the lock on `wt_die`, `SIGTERM`, normal exit. Identity-checked release (only unlinks if the lockfile still contains our PID — defensive). |
| Idempotent acquire | `WT_STATE_LOCK_HELD` guard so a function calling a helper that re-acquires doesn't deadlock on its own lock. |
| SIGKILL recovery | Manual `rm -f .worktrees.json.lock` once. Documented in the timeout error message. |

**No auto-steal.** Earlier iterations of this PR added stale-lock stealing for crash recovery (mkdir + holder file + age check + dead-PID check + atomic-mv steal + put-back). Under contention every variant we tried surfaced a phantom-acquire race where the steal+restore dance let two processes simultaneously believe they held the lock and race on `.worktrees.json`. Steal-free locks have no such window: every observer of the lockfile sees the same holder PID, no replacements happen behind any holder's back. SIGKILL'd holders leave an orphaned lockfile; the next acquirer's timeout fires with a clear recovery hint.

Wired into the three state-mutating paths:

- **`cmd_worktree_add`** — from `wt_init_state` through `wt_state_set`. Released BEFORE the optional `--start cmd_worktree_up` so a parallel add isn't blocked behind a docker compose up.
- **`cmd_worktree_remove`** — from after arg-parse through end-of-function, including the `read -rp` confirm prompt.
- **`cmd_worktree_set_env`** — from after env-validation through `wt_write_override`.

Read-only commands (`worktree list`, `regen`, `prune --dry-run`) intentionally left unlocked. `cmd_worktree_prune` mutates state and is a known follow-up.

## How we got here (commit history)

The investigation took a few iterations as race conditions surfaced under different stress patterns:

1. Initial: `mkdir(2)`-based lockdir with age-based stale-steal. CodeRabbit caught a TOCTOU on the steal/release path.
2. Owner-aware steal: PID-checked steal + identity-checked release. CodeRabbit caught a second-order race between two stealers.
3. Atomic-mv steal: atomic move + re-verify against steal criteria. Under heavy contention an instrumented trace showed a phantom-acquire — A's `mkdir` succeeded but C's mv stole A's lockdir before A could write the holder file. A returned "success" with no holder identity, and a concurrent B then also acquired.
4. `link(2)`-based atomic-with-content acquisition. Eliminated the mkdir+holder window. But steal+put-back retained tighter phantom windows under 16-core CPU load.
5. **Removed auto-steal entirely.** With no steal there's no race. Lockfile is created via `set -C` for simplicity (no staging file, pure shell). 100/100 stress under load.

## Tests

- **New file `state_lock.bats` (8 cases)**: acquire/release roundtrip, idempotent acquire, held-lock blocks second acquirer, ancient lock is NOT stolen (pins the no-steal contract), identity-checked release, no-op release, fail-fast on bad parent dir, concurrent acquirers serialize (lost-RMW canary).
- **`worktree_concurrent.bats`**: un-skipped the 3-parallel-adds test (was 5/5 red pre-fix, now deterministic green).

Version bump → 1.0.43.

## Local results

- `bats tests/bats/openemr-cmd/` — **134 ok, 0 failures**
- ShellCheck clean
- Concurrent lost-RMW test stressed 100x under 16-core CPU load: deterministic green
- Net script change: −273 lines (lots of steal-and-verify logic removed)

## Performance impact

Serializes the state-mutating phase of `add` / `remove` / `set-env`. For N concurrent agents adding worktrees, worst-case sequential time is N × ~10s (single add is fetch + git + file writes). Practical impact (2-3 agents): 20-30s — acceptable for the correctness gain.

## Paired follow-up

- openemr/openemr#12614 — adds `.worktrees.json.lock` to openemr's `.gitignore` so a SIGKILL-orphaned lockfile doesn't show up in `git status`.

## Test plan

- [x] Full local bats suite passes (134 ok)
- [x] ShellCheck clean
- [x] Concurrent test stress-runs 100/100 deterministic under 16-core CPU load
- [x] CI: BATS (ubuntu + macos), ShellCheck, e2e (real docker), CodeRabbit — all green

## Summary by CodeRabbit

* **New Features**
  * Added a state-lock mechanism to make concurrent worktree operations more reliable and prevent conflicting updates.
  * Updated the command version to include the latest improvements.

* **Bug Fixes**
  * Improved handling of simultaneous worktree add/remove/set-env actions to reduce race conditions and preserve state consistency.
  * Strengthened regression coverage for concurrent worktree operations and lock behavior.


bradymiller:
@coderabbitai review

bradymiller:
https://github.com/coderabbitai review

bradymiller:
@coderabbitai review

PR #844, opened 2026-06-26
ci(openemr-cmd): add macos-smoke-e2e on macos-14 via colima (daily-cron only)

bradymiller:
## ⚠️ Pre-merge cleanup required

The `if:` guard temporarily includes `pull_request` so this PR's CI can validate the job actually works end-to-end on a macos-14 runner (the alternative would be merging schedule-only blind). Before merge, drop the `pull_request` clause so the guard reverts to:

```yaml
if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
```

---

## Summary

New e2e job exercising one nonworktree-style stack (\`openemr-cmd up\` from \`docker/development-easy/\`) on a \`macos-14\` runner via colima. Pins two contracts on macOS Docker:

1. **openemr-cmd happy path works on macOS** (not just Linux Docker Engine)
2. **HOST_UID adoption holds through colima's Lima VM boundary** — \`sites/default/sqlconf.php\` should end up owned by the runner uid on the host (same probe as Linux \`nonworktree-lifecycle-e2e\`)

## Triggers

\`schedule\` + \`workflow_dispatch\` only. **Skipped on \`push\`/\`pull_request\`** via:
\`\`\`yaml
if: \${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
\`\`\`

Rationale: keeps PR cycle fast. macOS runners are limited in the GH-hosted pool and a stack startup on Apple Silicon + qemu amd64 emulation takes 15-25 min vs ~5 min on Linux. Catching macOS-specific divergence within 24h via the daily cron is the right tradeoff.

## Scope (intentionally narrow vs the Linux jobs)

| Linux | macOS smoke |
|---|---|
| worktree lifecycle | — |
| multi-concurrent (4 stacks) | — |
| nonworktree lifecycle | **✓ (this job)** |
| functional round-trip | — |
| prek install + real commit | — |

Just one stack, one health check, one HTTP smoke, one ownership assertion, one teardown. Heavier macOS-specific coverage can be filed later if this proves stable.

## Docker setup

\`\`\`yaml
- run: brew install colima docker docker-compose
- run: colima start --cpu 4 --memory 8 --disk 30
\`\`\`

macOS GH-hosted runners have brew preinstalled but no Docker. Apple's license forbids Docker Desktop on hosted runners; colima is unrestricted.

## Validation path

This PR can't validate the job end-to-end (it's gated to schedule/dispatch). After merge:
1. **GitHub Actions UI → openemr-cmd e2e workflow → "Run workflow" → branch=master** to verify the job works
2. **Tomorrow's daily cron** at 06:17 UTC will fire it automatically

If something breaks at runtime, the diagnostics step dumps docker ps, volume ls, colima status, and the last 500 lines of both openemr + mysql container logs.

## Test plan

- [ ] YAML validates (verified locally via Python yaml.safe_load)
- [ ] Job's \`if:\` guard correctly skips on this PR (visible in the Actions tab — the job should NOT run on this PR)
- [ ] Other 5 e2e jobs unaffected — additive change
- [ ] After merge: manual \`workflow_dispatch\` run reaches "OK: bind-mount file owned by uid=…" on macos-14

## Summary by CodeRabbit

* **Tests**
  * Added a new macOS smoke-test (tier-2) GitHub Actions job that runs a minimal `openemr-cmd up` flow on macOS 14 using a colima-provided Docker VM.
  * Provisions the VM with increased CPU/RAM/disk, waits for container health (longer timeout), and performs an HTTP smoke check on `localhost:8300`.
  * Verifies macOS bind-mount ownership via host UID checks, improves failure diagnostics, and ensures reliable teardown with `openemr-cmd down`.

### DIFF
--- a/.github/workflows/test-bats-openemr-cmd-real-docker.yml
+++ b/.github/workflows/test-bats-openemr-cmd-real-docker.yml
@@ -1714,7 +1714,11 @@ jobs:
     # If this job ever proves stable enough to gate PRs, the if-guard
     # below can drop the schedule-only restriction.
     name: e2e — macOS smoke (colima, daily-cron only)
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
+    # TEMPORARY: include pull_request so this PR can validate the job
+    # actually works on a macOS runner before merging. Revert to
+    # schedule-only (drop the pull_request clause) in the final
+    # pre-merge commit on this branch.
+    if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || github.event_name == 'pull_request' }}
     runs-on: macos-14
     # macOS colima + image pulls + slow VM disk = ~3-5x Linux. Inner
     # healthy poll budget is 30 min, outer job 60 min with margin for
--- a/.github/workflows/test-bats-openemr-cmd-real-docker.yml
+++ b/.github/workflows/test-bats-openemr-cmd-real-docker.yml
@@ -40,6 +40,11 @@ name: openemr-cmd e2e (real docker)
 #     installed hooks — rather than invoking the hook scripts by
 #     hand, so a regression in the shim, hooks-dir resolution, or
 #     cwd-aware container routing surfaces here.
+#   - macos-smoke-e2e: schedule + workflow_dispatch only (skipped on
+#     PR/push so cycle stays fast). One nonworktree-style stack on a
+#     macos-14 runner via colima — proves the openemr-cmd happy
+#     path + HOST_UID adoption work on macOS Docker (not just Linux).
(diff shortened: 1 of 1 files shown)


bradymiller:
Pivoting to arm64 Linux runners (ubuntu-24.04-arm) instead — see follow-up PR. The macOS approach hit a fundamental wall (free GH macos-14 can't run Linux VMs reliably; macos-13 Intel pool is being phased out + queues for 10+ min). arm64 Linux gives us the actual signal (arm-architecture image verification) without the macOS-VM brittleness, since Docker on Linux uses native containerization regardless of CPU arch.

PR #860, opened 2026-07-15
chore(aws): Express Plus to Ubuntu 24.04, fix recovery restore, release OpenEMR 8.2.0

stephenwaite:

The Express Plus CloudFormation templates launch an EOL Ubuntu 20.04 AMI, and on modern Ubuntu/Docker the disk design silently breaks: image layers land under /var/lib/containerd on the unencrypted 8GB root volume instead of the encrypted practice volume, filling the root disk and hanging the launch (found via a stack where MariaDB couldn't even exec its healthcheck).

The recovery template is additionally broken on master since the appliance migration: it passes an invalid `-e` flag to launch.sh and calls restore.sh at a path where it no longer exists. Finally, KMS DeletionPolicy: Delete contradicted the retained backups — deleting a stack scheduled the key's destruction, after which the retained S3 archives and final volume snapshots become permanently undecryptable. This PR resolves all of the above and releases OpenEMR 8.2.0 through the appliance.

All changes validated on us-east-1/t3.small (2026-07-15): fresh launch on Noble, backup chain (mariadb-backup + duplicity → S3), full recovery-stack restore including incremental replay, and a clean 8.2.0 install.

Note: the templates fetch launch.sh from master, so merging deploys 8.2.0 through all previously distributed template copies, not just new downloads.

- Templates: resolve Ubuntu 24.04 AMI via Canonical's public SSM parameter (replaces stale hardcoded focal RegionData map)
- Templates: relocate both docker and containerd stores to the encrypted practice volume (modern Docker splits storage across both; only docker was relocated)
- Templates: Noble compatibility — aws-cli via snap with /usr/local/bin symlink for cron PATH, snapd seed wait, pip --break-system-packages for cfn-bootstrap (PEP 668), drop removed apt --force-yes
- Templates: DeletionPolicy/UpdateReplacePolicy Retain on the KMS key so retained backups and snapshots outlive the stack; add a KMS alias named after the stack so retained keys are identifiable
- Templates: raise instance creation timeout to PT40M (Noble first-boot + image pulls exceed PT20M on t3.small)
- Recovery template: drop invalid `launch.sh -e` flag; call restore.sh at its post-migration path (both currently break recovery on master)
- restore.sh: report the archive chain's recovery point before restoring, fail fast if the duplicity pull fails instead of proceeding to the database restore, list staged database manifests
- Appliance: OpenEMR 7.0.4 → 8.2.0 (compose + README)


jesdynf:
@stephenwaite One issue -- both of those patched JSONs are actually the outputs of `stack.py` and the stackmaker, which was designed to let me keep the recovery document in lockstep with the launcher, but stack.py isn't being updated with the changes, the JSONs are getting directly patched.

The best thing to do would be to update stack.py to the current version (4.10.2, really?) via requirements, then patch stack.py to emit the changes made to the JSONs, then make sure that the syntax for any given element hasn't changed (and I'm sure it has)...

... but you could also make the argument that the time of those tools has passed and just delete them instead, maybe. It's been a while since I needed them, at any rate.

stephenwaite:
Validated deploy and recovery of a fully deleted stack from its retained, KMS-encrypted backups (full + incremental chain replay). Thanks for the tip on `stack.py`, should've asked you at the start :)

jesdynf (approved):
Looks good -- thanks for the armoring, and I really appreciate you going to the trouble to run the end-to-end testing.

PR #862, opened 2026-07-22
Express 820

jesdynf:

Updates AWS Express for 8.2.0


jesdynf:
Okay, I don't know why this says I'm murking 100 files. I think it's because it predates the docker migration to the primary repo and somehow the count is against the tree when I started and not the current state of play?

jesdynf:
I've spotchecked some of the files I'm allegedly deleting and they're not in upstream right now.

jesdynf:
@kojiromike @bradymiller I think this is safe to commit but I don't want to be That Guy. Do I need to unwind this somehow or is the diff just hallucinating? This is *supposed* to be a modest patch to the Express code.

jesdynf:
Somehow I needed to resync again and that brought the patch back into sanity.

PR #877, opened 2026-08-29
fix(openemr-cmd): prek dispatch fails loud when cwd's worktree stack is down

bradymiller:
## Summary

Discovered while shipping openemr/openemr#13761: pre-commit hooks silently skipped validation on a commit made from a worktree whose stack was down. The commit shipped un-validated shellcheck findings (SC2310, SC2312) that only surfaced later when I brought the stack up and manually ran prek.

## Root cause

The prek dispatch's cwd-aware routing has a fallback:

```bash
prek_target_id=$(wt_resolve_container_from_cwd)
[[ -n "${prek_target_id}" ]] || prek_target_id="${CONTAINER_ID}"
```

`wt_resolve_container_from_cwd` returns empty in two very different situations:
- cwd is NOT a worktree (primary repo or arbitrary dir) — fallback to `CONTAINER_ID` is correct
- cwd IS a worktree, but its stack is down — fallback is **wrong**: it routes prek into some OTHER worktree's container, which reads a different checkout via its own bind mount. Inside there, `git diff --staged --name-only` sees an empty index (that worktree's staged files, not ours). Every pre-commit hook reports "no files to check" and exits 0. The commit-side git operation then succeeds, and the operator sees a clean pass.

The current code cannot distinguish the two cases, so it always falls through — silently misrouting when a worktree stack happens to be down.

## Fix

`wt_resolve_container_from_cwd` now emits both the resolved branch AND container on stdout, tab-separated. Three output shapes distinguish the three states:

| Output | Meaning | Prek action |
|---|---|---|
| `<TAB>` | Not in a managed worktree | Fall back to `CONTAINER_ID` (unchanged) |
| `<branch><TAB>` | In this worktree; stack is DOWN | **Fail loud** with `worktree '<branch>' stack is not running` |
| `<branch><TAB><id>` | In this worktree; stack is UP | Dispatch to `<id>` |

Callers use `IFS=$'\t' read -r branch container < <(wt_resolve_container_from_cwd)` — process substitution, not `$()`, because command substitution runs in a subshell that discards any variable state (an earlier attempt using a `WT_RESOLVED_BRANCH` global didn't work for exactly this reason — flagged in the function docstring so the next person doesn't step on the same rake).

The fail-loud message names the specific worktree and points at the fix command:

```
Error: worktree 'my-branch' stack is not running.
       Start it with: openemr-cmd worktree up my-branch
       (Refusing to fall back to another running container — that would silently
        validate the wrong checkout and let the commit through un-validated.)
```

## Tests

New `tests/bats/openemr-cmd/prek_worktree_routing.bats` with 5 tests:

- 3 tests for `wt_resolve_container_from_cwd`'s tab-separated output contract (one per output shape)
- 2 tests for prek dispatch behavior:
  - cwd IS a worktree, stack DOWN → exit 1 naming the branch, refuses `CONTAINER_ID` fallback (asserted via absence of `docker exec` invocation in the recorded stub log)
  - cwd is NOT a worktree → falls back to `CONTAINER_ID` as before (unchanged behavior regression guard)

Each test uses a docker stub that toggles the label-filter ps output via `LABEL_PS_OUTPUT` env, plus a per-test git repo + fixture `.worktrees.json`.

## Verification

- New file: 5/5 tests pass
- Existing `prek_dispatch.bats` + `container_resolution_worktree_fallback.bats`: 9/9 tests still pass (no regression)
- Full `tests/bats/openemr-cmd/` suite: 313/317 pass; 4 failures are pre-existing "remove probe" tests that fail when running as root in the container (unrelated to this change)

### DIFF
--- a/tests/bats/openemr-cmd/prek_worktree_routing.bats
+++ b/tests/bats/openemr-cmd/prek_worktree_routing.bats
@@ -1,0 +1,232 @@
+# BATS: `openemr-cmd prek` — cwd-aware worktree container routing.
+#
+# When git commits fire from inside a managed worktree, the pre-commit
+# git hook invokes `openemr-cmd prek run`, and prek must dispatch to
+# THAT worktree's container — never to some other worktree's container
+# just because it happens to be running. Silent misrouting was the bug:
+# the pre-commit run would exec inside the wrong container, which reads
+# a different checkout via its own bind mount and reports "no files to
+# check" for every hook (its git index doesn't match this worktree's
+# staged files). The commit then completed with zero validation.
+#
+# Fix: `wt_resolve_container_from_cwd` now emits `<branch><TAB><container>`
+# on stdout, letting the prek dispatch distinguish the three states via
+# the tab-separated pair:
+#
+#   `<TAB>`               cwd is NOT a worktree (primary repo or
+#                         arbitrary dir). Fallback to CONTAINER_ID is
+#                         correct.
+#   `<branch><TAB>`       cwd IS a worktree but its stack is down.
+#                         Refuse to fall back; error with the worktree
+#                         name so the operator knows exactly which
+#                         stack to bring up.
+#   `<branch><TAB><id>`   Happy path — dispatch to `<id>`.
+#
+# Complements prek_dispatch.bats (which pins the outer docker exec
+# shape via -d bypass, sidestepping cwd routing entirely).
+
+load '../test_helper/bats-support/load'
+load '../test_helper/bats-assert/load'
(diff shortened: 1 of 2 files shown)


kojiromike:
What does "stack is down" mean in this context?

bradymiller:
> What does "stack is down" mean in this context?

hi @kojiromike ,
Clarified echoed instructions that "stack is down" can either mean the worktree docker stack is stopped or does not exist. The bug was that if this was the case, then the precommit was silently running through another worktree stack (rather than failing, which it now does).

PR #881, opened 2026-09-14
chore(release): 8.4.0 release

jesdynf:

8.4.0 release updates to scripts
-
-
-


bradymiller:
hi @jesdynf , Would it be helpful to automate this stuff on releases (i could trigger this to happen since also need to make the change on the kubernetes package also). Wouldn't be tough.

